PR #2, opened 2018-05-23
Fix incorrect spelling of 'than'

superkarolis:
Hey, just a small PR. Impressed with the project, will use it!


trekhleb:
Thank you for pull request @superkarolis! Great that you've noticed the type.
Looks like tests failed. Probably lessThan has not been fixed in all places.
In order to still have this fix I made it manually and commit. So the typo is fixed now.

superkarolis:
My mistake, thanks!

PR #3, opened 2018-05-23
small refactor in some sorting algorithms for better readability

PeterShershov:
Small change to make it more readable using the spread operator.


trekhleb:
@PeterShershov thank you for your fix! Spread operator you've used is better for readability of course.

PR #8, opened 2018-05-23
Fixed Typo in Big O Complexity table

breinhart:
Just a typo fix in the README.


trekhleb:
@breinhart  thank you for noticing and fixing that!

PR #11, opened 2018-05-24
1 is not prime

ricsam:


trekhleb:
@ricsam thank you for pointing that out! 
The tests for current pull request have been failed though. 
So I've fixed the issue manually.

PR #14, opened 2018-05-24
Fix a typo

TomasHubelbauer:


trekhleb:
Thank you for the fix @TomasHubelbauer

PR #18, opened 2018-05-25
Finish main part of translation

mybaseball52:
Finish main part of translation


trekhleb:
@mybaseball52 thank you again for submission!

What about algorithm names like "Bridges", "Traveling Salesman Problem" and "Knapsack Problem" and so on? Could you please translate all of them as well so that translation file would be complete?

trekhleb:
Great! Thank you for translation!

PR #20, opened 2018-05-25
Added Linear Search

ak4522912:
Added algorithm for the basic and useful linear search


trekhleb:
@ak4522912 thank you for submission!

Could you please fix code style issues? You can find what issues exactly are there by running `npm run ci` locally.

PR #27, opened 2018-05-28
Use Map for TrieNode children

caub:
`Map` are more suitable and efficient for dynamically shaped hashes ([quick benchmark](https://repl.it/@caub/obj-vs-map-perf))


trekhleb:
@caub 
Thank you for submission! The code looks much cleaner with using Map() and that is nice.
But could you please revert your changes in package-lock.json? Seems like they are not relevant, aren't they?

caub:
done
There's one thing slower with Map though, it's `[...map.keys()]` vs `Object.keys(obj)`, but I think `.suggestChildren` is not used too often

trekhleb:
Great! Thank you for contribution!

PR #33, opened 2018-05-30
Fix remove method in AvlTree: Nodes can not re-balance.

Bladefidz:
Remove node in AvlTree with auto balancing.

Fix issue: trekhleb#13


trekhleb:
Wow, @Bladefidz , thank you for your huge effort! Amount of code is impressive!

I've already pushed Red-Black-Tree implementation though. The main difference between the code in master branch and this PR is that in master the Red-Black tree is based upon Binary Search tree. I think it is good approach to prevent code duplication and also because Red-Black trees are BSTs by definition. 

Could you please re-target this pull request so that it purpose would be balancing AVL tree while removing the node from it?

Bladefidz:
Okay @trekhleb, I will modify my PR so only commits on AVLTree will be pulled.

trekhleb:
@Bladefidz thank you for PR!

PR #36, opened 2018-05-30
fibonacciNth function added

ohepworthbell:
The `fibonacci()` function now returns array of the fibonacci sequence, with the function `fibonacciNth()` returning the value at nth-position in the fibonacci sequence.


trekhleb:
@ohepworthbell thank you for cleaning the code up and for clean implementation of new function!

PR #42, opened 2018-06-01
Fix LinkedList

m-maksyutin:
This is a small fix for LinkedList `delete` method. I have updated LinkedList.test.js to reproduce the issue. This test fails for the previous version of the method.


Trapperjs:
Is there a secret back door I could use?

PR #50, opened 2018-06-04
 Fix the remove method for the MinHeap

m-maksyutin:
Previously the custom finding comparator was not fully used in the remove method.


trekhleb:
Thank you for the fix @m-maksyutin !

PR #52, opened 2018-06-04
add recursive solution to permutations with repetitions problem

mmapplebeck:


trekhleb:
Thank you for submission @mmapplebeck

PR #55, opened 2018-06-07
Fix the translation of sorting complexity table

team6612:
Array "Sorting" Algorithms Complexity should be 陣列排序演算法, not 陣列搜尋演算法


trekhleb:
Thank you for contribution @team6612 !

PR #60, opened 2018-06-10
Fix bug in 'BinarySearchTreeNode.js

rakshithjk:
I think that properties of the returned object correspond to the tree state before the node is deleted.

``` 110.         nodeToRemove.value = nodeToRemove.right.value;```

You have changed the value of ``` nodeToRemove```  with the value that has to be replaced.Now when you return ``` nodeToRemove.value```  it would return the node that is being replaced to that position, not the node that is being deleted. Instead return ``` parent```


trekhleb:
Thank you for PR @rakshithjk 
I've already made changes. Instead of removed node the removal method now will return just boolean to avoid confusion.

PR #61, opened 2018-06-11
feat: Add Liu Hui's π algorithm

mantou132:


trekhleb:
Thank you for contribution @mantou132

PR #62, opened 2018-06-11
docs: update correct Big-O chart

tiendq:
Current Big-O graph in README files are not same with the original image on [bigocheatsheet.com](http://bigocheatsheet.com/). O(n) is fair, not good.


trekhleb:
Thank you for the fix @tiendq

PR #67, opened 2018-06-13
added Bidirectional Conditional Insertion Sort

christian4423:
This is a Bidirectional Conditional Sort. 

Note on stability with custom compare function. 

```
let comp = new Comparator(function (a, b) {
    let x = a;
    let y = b;
    if (typeof a === "string") {
        x = a.length;
    }
    if (typeof b === "string") {
        y = b.length;
    }
    if (x === y) {
        return 0;
    }
    return x < y ? -1 : 1;
})
```

i.e. 
**Without**
a, cc, bbb --> a, bbb, cc
**With**
a, cc, bbb --> a, cc, bbb

Even with the custom compare, you are not always guaranteed a stable sort
Example: Given: [aa, q, a, bbbb, ccc] 
The algorithm will start with a SWAP from the right most index and the middle of the array.
1. START SWAP "ccc", "a"
END SWAP (New Array: aa,q,ccc,bbbb,a)
2. "aa" is greater than "a"
3. START SWAP "aa", "a"
END SWAP a,q,ccc,bbbb,aa
4. "ccc" is greater than "aa"
5. START INSRIGHT  (Given: a,q,bbbb,bbbb,aa)
END INSRIGHT (New Array: a,q,bbbb,aa,ccc)
6. "bbbb"is greater than "aa"
7. START INSRIGHT (Given: a,q,bbbb,aa,ccc)
END INSRIGHT  (New Array: a,q,aa,ccc,bbbb)

Finished:
[ 'a', 'q', 'aa', 'ccc', 'bbbb' ]


christian4423:
Going to make a new one.

PR #71, opened 2018-06-16
Permutation and Combination Calculation

amitsingh19975:
It will help in calculation of total number permutation and combination possible for elements.


trekhleb:
Thank you for your PR @amitsingh19975 

There is already Permutation and Combinations sections in this repo. Please check this links:
- https://github.com/trekhleb/javascript-algorithms/tree/master/src/algorithms/sets/permutations
- https://github.com/trekhleb/javascript-algorithms/tree/master/src/algorithms/sets/combinations

So could you please create a new PR only for complex numbers.
Also please avoid using `*.prototype` but rather use ES6 syntax in classes.

amitsingh19975:
OK thanks

Issue #63, opened 2018-06-12
knapsack item settings ignores itemsInStock


shSina:
the algorithm ignores itemsInStock count

shSina:
the Unbounded knapsack problem ignores the total number of itemsInStock
```js
const possibleKnapsackItems = [
  new KnapsackItem({ value: 1, weight: 1, itemsInStock: 10 }),
  new KnapsackItem({ value: 100, weight: 100, itemsInStock: 100 }),
  new KnapsackItem({ value: 50, weight: 50, itemsInStock: 2 }),
  new KnapsackItem({ value: 1000, weight: 1000, itemsInStock: 5 }),
];
const maxKnapsackWeight = 76;

const knapsack = new Knapsack(possibleKnapsackItems, maxKnapsackWeight);

knapsack.solveUnboundedKnapsackProblem();

console.log(knapsack.totalValue);
console.log(knapsack.totalWeight);
console.log(knapsack.selectedItems.length);
console.log(knapsack.selectedItems[0]);
console.log(knapsack.selectedItems[1]);
76
76
2
KnapsackItem { value: 1, weight: 1, itemsInStock: 10, quantity: 26 } 
KnapsackItem { value: 50, weight: 50, itemsInStock: 2, quantity: 1 } 
```

shSina:
will fix as below , I will make e pr:
```js
solveUnboundedKnapsackProblem() {
    this.sortPossibleItemsByValue();
    this.sortPossibleItemsByValuePerWeightRatio();

    for (let itemIndex = 0; itemIndex < this.possibleItems.length; itemIndex += 1) {
      if (this.totalWeight < this.weightLimit) {
        const currentItem = this.possibleItems[itemIndex];

        // Detect how much of current items we can push to knapsack.
        // if(currentItem.quantity == currentItem.itemsInStock) return;
        const availableWeight = this.weightLimit - this.totalWeight;
        const maxPossibleItemsCount = Math.floor(availableWeight / currentItem.weight);

        if (maxPossibleItemsCount) {
          if(maxPossibleItemsCount <= currentItem.itemsInStock){
            currentItem.quantity = maxPossibleItemsCount;
          }else{
            currentItem.quantity = currentItem.itemsInStock;
          }
          this.selectedItems.push(currentItem);
        }
      }
    }
  }
```

trekhleb:
Nice catch @shSina. Waiting for PR.

trekhleb:
[...]

PR #73, opened 2018-06-17
fix unbound knapsack problem with items more than 1(default value)

shSina:
fix bug #63


trekhleb:
Thank you for the fix @shSina!

PR #74, opened 2018-06-19
Fix BST removal method

m-maksyutin:
There was an issue with the parent node. Now the removal method works for nodes with no parent.


trekhleb:

PR #75, opened 2018-06-19
Corrected explanations and included an example

Nnadozie:


trekhleb:
Thank you for your PR @Nnadozie

PR #80, opened 2018-06-25
 Fix the findEdge method of the graph

m-maksyutin:


trekhleb:
Thanks for another fix, @m-maksyutin ! :)

PR #81, opened 2018-06-25
Fix the value returned by DisjointSet union

m-maksyutin:
Minor fix to match the method documentation


trekhleb:

PR #82, opened 2018-06-26
Fix the prepend method

SimoTw:
Add the newNode.next reference , so that the new head(newNode) and the old head(this.head before giving new reference value ) will be linked by .next after prepend method.


m-maksyutin:
Hi @SimoTw , it seems that LinkedListNode constructor does it.

SimoTw:
@m-maksyutin 
your right, sorry didn't aware it, thanks.

PR #86, opened 2018-07-01
Fix the insertion sort

m-maksyutin:
Hi, this is fixing the situation when we have an unsorted array with negative numbers and zeroes.


trekhleb:

PR #92, opened 2018-07-04
Added doubly linked list

huzaima:
Doubly linked list with the same methods that are used in singly linked list. Tests are also added.


trekhleb:
Thank you for your effort @huzaima

PR #93, opened 2018-07-05
Fix some errors in README.md 

Kazexiii:
- Simplify some sentences in Stack, Tree and Trie
- Clarification of concepts in Stack, Tree and Trie

### DIFF
--- a/src/data-structures/stack/README.md
+++ b/src/data-structures/stack/README.md
@@ -3,12 +3,12 @@
 In computer science, a stack is an abstract data type that serves 
 as a collection of elements, with two principal operations:
 
-* **push**, which adds an element to the collection, and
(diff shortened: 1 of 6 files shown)


trekhleb:
Thank you for your PR @Kazexiii!

But it seems for me that most of the changes are just matter of the taste. For example:

- changing "_... the key with which it is associated_" to "_... associated key_"
- changing "_Rather, values tend only to be associated with leaves_" to "_Rather, values tend to be associated with leaves_".

It looks like both versions are correct and several people may have different taste and may create similar PRs to change it back and forth.

It would be really nice to have text errors to be fixed instead. Or to have some additional value to be added in form of new paragraphs or explanation points.

PR #99, opened 2018-07-20
Update README.md

knicholes:


trekhleb:
Hi @knicholes,

Thank you for your PR. "Data Structure Operations Complexity" table contains worst case scenarios for basic DS operations. This is how big O complexity is being calculated. Thus there is no need to mention edge cases like with Sorted Array.

PR #100, opened 2018-07-20
Fix CountingSort for negative numbers

dubzzz:
When trying to run property based tests on the sorting algorithms available in javascript-algorithms, I detected sorting issues on CountingSort and RadixSort with arrays having negative entries.

Here is a fix to be able to handle negative integers in CountingSort.

The commit adding property based tests to the sorting algorithms is available at https://github.com/dubzzz/javascript-algorithms/commit/34b0c46df41fc18ab839be8c8bdece556d3b5d70
If you are interested in such testing approach I can issue another pull request to add those tests too.


trekhleb:
Thank you for your PR @dubzzz

PR #101, opened 2018-07-20
Fix knuthMorrisPratt for empty word request

dubzzz:
While running property based tests - *based on fast-check* - on the implementation of knuthMorrisPratt I discovered what seems to be an inconsistency:

```javascript
knuthMorrisPratt("", "") === -1
knuthMorrisPratt("a", "a") === 0
```

The tests that discovered the issue are available in the commit: https://github.com/dubzzz/javascript-algorithms/commit/87329dd9b0215683018a346694cc2bf387fe32b4
If you are interested in this test method, I can issue another pull request for those property based tests.

### DIFF
--- a/src/algorithms/string/knuth-morris-pratt/__test__/knuthMorrisPratt.test.js
+++ b/src/algorithms/string/knuth-morris-pratt/__test__/knuthMorrisPratt.test.js
@@ -2,6 +2,9 @@ import knuthMorrisPratt from '../knuthMorrisPratt';
 
 describe('knuthMorrisPratt', () => {
   it('should find word position in given text', () => {
+    expect(knuthMorrisPratt('', '')).toBe(0);
+    expect(knuthMorrisPratt('a', '')).toBe(0);
+    expect(knuthMorrisPratt('a', 'a')).toBe(0);
     expect(knuthMorrisPratt('abcbcglx', 'abca')).toBe(-1);
     expect(knuthMorrisPratt('abcbcglx', 'bcgl')).toBe(3);
     expect(knuthMorrisPratt('abcxabcdabxabcdabcdabcy', 'abcdabcy')).toBe(15);
--- a/src/algorithms/string/knuth-morris-pratt/knuthMorrisPratt.js
+++ b/src/algorithms/string/knuth-morris-pratt/knuthMorrisPratt.js
@@ -30,6 +30,10 @@ function buildPatternTable(word) {
  * @return {number}
  */
 export default function knuthMorrisPratt(text, word) {
+  if (word.length === 0) {
+    return 0;
+  }
+
   let textIndex = 0;
   let wordIndex = 0;
 


src/algorithms/string/knuth-morris-pratt/__test__/knuthMorrisPratt.test.js
```diff
@@ -2,6 +2,9 @@ import knuthMorrisPratt from '../knuthMorrisPratt';
 
 describe('knuthMorrisPratt', () => {
   it('should find word position in given text', () => {
+    expect(knuthMorrisPratt('', '')).toBe(0);
+    expect(knuthMorrisPratt('a', '')).toBe(0);
+    expect(knuthMorrisPratt('a', 'a')).toBe(0);
```
trekhleb: This test is already valid without main function code changes.

src/algorithms/string/knuth-morris-pratt/__test__/knuthMorrisPratt.test.js
```diff
@@ -2,6 +2,9 @@ import knuthMorrisPratt from '../knuthMorrisPratt';
 
 describe('knuthMorrisPratt', () => {
   it('should find word position in given text', () => {
+    expect(knuthMorrisPratt('', '')).toBe(0);
+    expect(knuthMorrisPratt('a', '')).toBe(0);
```
trekhleb: This test is actually not correct. It says that when I search for an empty word '' in a string 'a' I should find one and this empty word will be accessible at zero index. But `string[0]` will return us `a` which expected but is not correct according to this test. So I assume that we should expect `-1` here instead of `0`.

src/algorithms/string/knuth-morris-pratt/__test__/knuthMorrisPratt.test.js
```diff
@@ -2,6 +2,9 @@ import knuthMorrisPratt from '../knuthMorrisPratt';
 
 describe('knuthMorrisPratt', () => {
   it('should find word position in given text', () => {
+    expect(knuthMorrisPratt('', '')).toBe(0);
```
trekhleb: I think this test is wrong. We should expect `-1` here instead of `0` because having zero as a result would mean that I may fetch empty word out of `string[0]`. But this is not true since the `string` itself is empty.

src/algorithms/string/knuth-morris-pratt/knuthMorrisPratt.js
```diff
@@ -30,6 +30,10 @@ function buildPatternTable(word) {
  * @return {number}
  */
 export default function knuthMorrisPratt(text, word) {
+  if (word.length === 0) {
```
trekhleb: Please take a look at the comments to the tests above. No need to do this checking.

trekhleb:
@dubzzz thank you for you PR. Could you please check my comments in code to this PR though. It seems like tests are correct now. It is ok to receive `-1` when you're looking for an empty word in empty string. This would mean that there is no way of fetching empty word from empty string (`string[0] <-- exception, since string length is 0`).

src/algorithms/string/knuth-morris-pratt/__test__/knuthMorrisPratt.test.js
dubzzz: @trekhleb Actually I was understanding it a bit differently.

In my understanding I was expecting `knuthMorrisPratt` to return the same as `String.prototype.indexOf` - *more a less in a sense*.
In other word I considered the output as being: *index within `text` of the first occurrence of `word`* [[official reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/indexOf)].

```js
// definition in terms of code
const startIdx = knuthMorrisPratt(text, word);
text.substr(startIdx, word.length) === word // should be true for startIdx !== -1
```

Actually both `"".indexOf("") === 0` and `"a".indexOf("") === 0`.

dubzzz:
Actually both `String.prototype.indexOf` and `rabinKarp` agree on the expected results. I really think that the implementation of `knuthMorrisPratt` should be updated accorindly:

```js
console.log(rabinKarp('', '')); //0
console.log(rabinKarp('a', '')); //0
console.log(rabinKarp('a', 'a')); //0

console.log(''.indexOf('')); //0
console.log('a'.indexOf('')); //0
console.log('a'.indexOf('a')); //0
```

I can re-open the PR if you agree on it ;)

PR #109, opened 2018-07-24
 Translate to Korean

yeonjuan:
Translate to Korean.


trekhleb:
Awesome! Thanks for translation @yeonjuan

Issue #102, opened 2018-07-21
rabinKarp seems to miss some matches


dubzzz:
Hello,

I recently tried to run some property based tests on the algorithms provided in javascript-algorithms.
I discovered the following failing case with rabinKarp algorithm:

```javascript
expect(rabinKarp("^ !/'#'pp", " !/'#'pp")).toBe(2); // output: -1
```

In order to find it, I used [fast-check framework](https://github.com/dubzzz/fast-check/) with the property:

```javascript
// for any strings a, b and c
// rabinKarp(a + b + c, b) should be different from -1 (indeed b is a substring of a+b+c)
fc.assert(
  fc.property(
    fc.string(), fc.string(), fc.string(),
    (a, b, c) => rabinKarp(a + b + c, b) !== -1
  )
);
```

I was not able to have a smaller example for this precise case :/

dubzzz:
It seems to be some kind of overflow of `hashWord`/`reHashWord`. With previous strings the computed hash is above `Number.MAX_SAFE_INTEGER`.

```javascript
Number.MAX_SAFE_INTEGER; //9007199254740991
hashWord(" !/'#'pp"); //9143038766935264
reHashWord(hashWord("^ !/'#'p"), "^ !/'#'p", " !/'#'pp"); //9143038766935266
```

dubzzz:
Given the fact that the issue is coming from `word[charIndex].charCodeAt(0) * PRIME ** charIndex` going above `Number.MAX_SAFE_INTEGER`, it means we are safe up to `charIndex == 4` because `0x10ffff * 97 ** 4 < Number.MAX_SAFE_INTEGER` - false for 5.

The current implementation of the algorithm should be ok - *will confirm my test passes with this limit* - for any word having at most 5 characters otherwise there is no guarantee that it does not overflow and goes into rounding issues.

trekhleb:
Good catch @dubzzz

dubzzz:
@trekhleb Would you agree on a PR adding some property based tests in this repository?

All the issues I reported so far were found using this test method. For the moment I only added them for sorting, knuth and rabin in my local fork. I think I can spend some time to check the other algorithms too.

In a nutshell:

It is a testing approach complementary to examples based tests. The idea is to check for properties:
*for all (x, y, ...) such as precondition(x, y, ...) holds property(x, y, ...) is true*
The framework is then responsible to generate x,y..., such as precondition holds, run the check on them and build a smaller failing case in case of failure.

[...]

PR #110, opened 2018-07-24
Refactor Rabin-Karp

Bruce-Feldman:
Created Rabin hash function module.
Transitioned Rabin-Karp string matching function to Rabin hash function family.

### DIFF
--- a/src/algorithms/string/rabin-karp/__test__/rabinKarp.test.js
+++ b/src/algorithms/string/rabin-karp/__test__/rabinKarp.test.js
@@ -1,24 +1,20 @@
-import { rabinKarp, hashWord, reHashWord } from '../rabinKarp';
+import rabinKarp from '../rabinKarp';
 
 describe('rabinKarp', () => {
-  it('should correctly calculates hash and re-hash', () => {
-    expect(hashWord('a')).toBe(97);
-    expect(hashWord('b')).toBe(98);
-    expect(hashWord('abc')).toBe(941094);
-    expect(hashWord('bcd')).toBe(950601);
-    expect(reHashWord(hashWord('abc'), 'abc', 'bcd')).toBe(950601);
-    expect(reHashWord(hashWord('abc'), 'abc', 'bcd')).toBe(hashWord('bcd'));
-  });
-
   it('should find substring in a string', () => {
     expect(rabinKarp('', '')).toBe(0);
     expect(rabinKarp('a', '')).toBe(0);
     expect(rabinKarp('a', 'a')).toBe(0);
+    expect(rabinKarp('ab', 'b')).toBe(1);
     expect(rabinKarp('abcbcglx', 'abca')).toBe(-1);
     expect(rabinKarp('abcbcglx', 'bcgl')).toBe(3);
     expect(rabinKarp('abcxabcdabxabcdabcdabcy', 'abcdabcy')).toBe(15);
     expect(rabinKarp('abcxabcdabxabcdabcdabcy', 'abcdabca')).toBe(-1);
     expect(rabinKarp('abcxabcdabxaabcdabcabcdabcdabcy', 'abcdabca')).toBe(12);
     expect(rabinKarp('abcxabcdabxaabaabaaaabcdabcdabcy', 'aabaabaaa')).toBe(11);
+    expect(rabinKarp('^ !/\'#\'pp', ' !/\'#\'pp')).toBe(1);
+    expect(rabinKarp('a\u{ffff}', '\u{ffff}')).toBe(1);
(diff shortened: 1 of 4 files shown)


Bruce-Feldman:
I additionally think that Rabin-Karp should be changed to be a beginner topic.

dubzzz:
Actually it might fix #102 as the hash function has been changed.

dubzzz:
It seems that some values are still failing with this implementation. The characters outside the BMP-plan of Unicode make the algorithm fails:

```js
rabinKarp("a\u{ffff}", "\u{ffff}"); // => OK - return: 1
rabinKarp("a\u{10000}", "\u{10000}"); // => FAIL - return: -1

// \u{10000} is LINEAR B SYLLABLE B008 A
// more at https://www.fileformat.info/info/unicode/char/10000/index.htm
```

See: https://runkit.com/dubzzz/rabin-2

Bruce-Feldman:
This is a really cool bug.  For some reason, on my machine, I get the following interesting result:
```
> var text = "a\u{10000}";
undefined
> text.length
3
> Array.from(text).length
2
```

I am not so familiar with the guarantees of these library functions, is this expected behaviour?  With your first example, I get the following behaviour (which I would expect):
```
> var text = "a\u{1000}"
undefined
> text.length
2
> Array.from(text).length
2
```

dubzzz:
This is an expected behavior. In JavaScript, strings are encoded using UTF-16 which encodes code units on 16 bits. Actually characters outside the BMP range, which means whose code point is greater than 0xffff, require two code units to encode one code point. Legacy methods of JavaScript are counting code units while modern ones (most of the time) work at code point level (spread operator, Array.from...).

dubzzz:
If your code uses both `s.length` and `Array.from(s)` you must definitely choose one.

`s.length` for code point would be `[...s].length` or `Array.from(s).length`

`Array.from(s)` for code units would be `s. split('')`

dubzzz:
Just an additional note:

```js
'\u{10000}'.charCodeAt(0) // 0xd800 
'\u{10000}'.codePointAt(0) // 0x10000
```

Bruce-Feldman:
@dubzzz I updated the string processing to handle this edge case.  Could you take a look at it and see if it seems sufficient?  I think I understood your what you were saying about javascript representation of strings, but am not sure.

dubzzz:
@Bruce-Feldman Just retried based on the last code you pushed. It seems that there is still an issue:

```js
// your implementation
console.log(rabinKarp("a耀a","耀a")); // 1
console.log(rabinKarp("\u0000耀\u0000","耀\u0000")); // -1 ERROR

// indexOf
console.log("a耀a".indexOf("耀a")); // 1
console.log("\u0000耀\u0000".indexOf("耀\u0000")); // 1
```

You can easily re-run my test case either by using RunKit or locally by adding fast-check package.
Snippet: https://runkit.com/dubzzz/rabin-3 (tell me if you can't access it)

Bruce-Feldman:
Any recommendations to fix this problem in the general case?

Bruce-Feldman:
@dubzzz Looks like the issue was that I forgot that javascript performs bit operations on signed 32 bit integers.  This should be addressed now by using non bit operators.

dubzzz:
With your last implementation, all my property based tests are green ;)

On my side, I think we should integrate such property based testing tools into the repository to have more confidence in the algorithms and cover all possible and legits corner cases. In the case of `rabinKarb` it helped to discover multiple issues: in the current and the new implementation.

Concerning general ways to cover those edge cases on character encodings I do not have real recommendations :/

Bruce-Feldman:
@dubzzz Woohoo!  I am really pleased with your tests - they seem to generate edge cases fairly reliably.  Thanks for the help!

trekhleb:
@Bruce-Feldman, @dubzzz thank you for PR and for new Rabing Karp testing edge cases!

@Bruce-Feldman I like the idea of moving hashing functionality out of RabinKarp. I even think that it should be moved not in `utils` but into separate section that will contain different hash implementations and maybe different crypto related algorithms. I'll merge your PR to separate branch for now and create this new section for hashing with related READMEs and links for further readings.

@dubzzz regarding the tool you've created that does property based testing - it is really nice I guess since it helped to find such edge cases for Rabing Karp as `^ !/\'#\'pp` vs ` !/\'#\'pp'`. But currently I'm not ready to answer your question about whether we'll have it in current repo or not. Let me investigate it further since I didn't have a chance to play with it so far.

dubzzz:
@trekhleb Thanks for the update. Please keep me updated if you need more details on the approach. Meanwhile I will try to find some more times to checks other algorithms

trekhleb:
Thank you @dubzzz

PR #112, opened 2018-07-25
Add rain-terraces (trapping rain water) algorithm

KeoniGarner:
Should probably be a basic/beginner algorithm


trekhleb:
@KeoniGarner thank you for your PR. It looks like the code has linting issues. Could you please fix them so that tests are passed? You may lint code by running `npm run lint` locally.

KeoniGarner:
Thank you for the quick response! I think that should take care of the linting.

PR #113, opened 2018-07-25
Refactored disjoint set data structure

Bruce-Feldman:
Refactored disjoint set data structure
Updated dependencies (Kruskal, Detect Cycles) for new interface.


Bruce-Feldman:
I also think this data structure should be marked as beginner.  The only non-trivial piece is the path halving portion, which is fairly straightforward once explained.

src/algorithms/graph/detect-cycle/detectUndirectedCycleUsingDisjointSet.js
```diff
@@ -6,26 +6,19 @@ import DisjointSet from '../../../data-structures/disjoint-set/DisjointSet';
  * @param {Graph} graph
  */
 export default function detectUndirectedCycleUsingDisjointSet(graph) {
-  // Create initial singleton disjoint sets for each graph vertex.
-  /** @param {GraphVertex} graphVertex */
-  const keyExtractor = graphVertex => graphVertex.getKey();
-  const disjointSet = new DisjointSet(keyExtractor);
-  graph.getAllVertices().forEach(graphVertex => disjointSet.makeSet(graphVertex));
+  const vertices = graph.getAllVertices();
+  const disjointSet = new DisjointSet(vertices.length);
```
trekhleb: I think it would be better not to lock Disjoint Set size from the very beginning but make it available to grow instead. Thus one of the basic Disjoint Set operation as "MakeSet" becomes possible.

src/algorithms/graph/detect-cycle/detectUndirectedCycleUsingDisjointSet.js
```diff
@@ -6,26 +6,19 @@ import DisjointSet from '../../../data-structures/disjoint-set/DisjointSet';
...
-  // edge will create a cycle.
-  let cycleFound = false;
-  /** @param {GraphEdge} graphEdge */
-  graph.getAllEdges().forEach((graphEdge) => {
-    if (disjointSet.inSameSet(graphEdge.startVertex, graphEdge.endVertex)) {
-      // Cycle found.
-      cycleFound = true;
-    } else {
-      disjointSet.union(graphEdge.startVertex, graphEdge.endVertex);
+  for (let i = 0; i < vertices.length; i += 1) {
```
trekhleb: I still prefer object-oriented manner of dealing with set elements instead of index-based. Just for learning and code reading purposes.

Take a look ta these two examples:
```javascript
graph.getAllEdges().forEach((graphEdge) => {
  if (disjointSet.inSameSet(graphEdge.startVertex, graphEdge.endVertex)) {
    ...
  }
}
```

and 

```
for (let i = 0; i < vertices.length; i += 1) {
  for (let j = (i + 1); j < vertices.length; j += 1) {
    if (vertices[i].hasNeighbor(vertices[j])) {
      if (disjointSet.find(i) === disjointSet.find(j)) {
        ...
      }
    }
  }
}
```

It feels like the first one is more readable.

src/algorithms/graph/detect-cycle/detectUndirectedCycleUsingDisjointSet.js
```diff
@@ -6,26 +6,19 @@ import DisjointSet from '../../../data-structures/disjoint-set/DisjointSet';
...
+    for (let j = (i + 1); j < vertices.length; j += 1) {
+      if (vertices[i].hasNeighbor(vertices[j])) {
+        if (disjointSet.find(i) === disjointSet.find(j)) {
+          return true;
+        }
+        disjointSet.union(i, j);
```
trekhleb: Object-oriented version of union has better readability as for me as well comparing to index-based. Take a look:

```javascript
disjointSet.union(graphEdge.startVertex, graphEdge.endVertex);
```

vs

```javascript
disjointSet.union(i, j);
```

src/data-structures/disjoint-set/DisjointSet.js
```diff
@@ -1,97 +1,46 @@
-import DisjointSetItem from './DisjointSetItem';
-
 export default class DisjointSet {
   /**
-   * @param {function(value: *)} [keyCallback]
+   * @param {number} [size] - Number of disjoint elements in set
    */
-  constructor(keyCallback) {
```
trekhleb: It is really good to have a possibility to put any objects in Disjoint Set. If we won't have this `keyCallback` it won't be possible. I understand that it makes Disjoint Set implementation more verbose but from the other side it makes the usage of Disjoint Set class less verbose and more convenient since I don't need to work with indices but rather with objects directly.

trekhleb:
@Bruce-Feldman thank you for your PR. I would prefer to still have this object-oriented approach in Disjoint Set realization instead of index-based. I understand that it makes Disjoint Set implementation less verbose and bulky. But I believe it make the consumption of Disjoint Set class to be more pleasant and make the code more readable. Also it seems really nice to have Disjoint Set being resizable. I've described this in more details in comments to the code above. Anyways thank you for your effort.

src/algorithms/graph/detect-cycle/detectUndirectedCycleUsingDisjointSet.js
Bruce-Feldman: This seems reasonable & is easy to implement in the current scheme.

src/algorithms/graph/detect-cycle/detectUndirectedCycleUsingDisjointSet.js
Bruce-Feldman: I do not think that this is a reasonable comparison.  Please let me know if I am wrong, but it very much seemed to me that the implementation of disjoint set was extremely influenced by the current implementation of graph.  I didn't think this was appropriate, as they are two separate data structures whose various implementations should be clear and distinct.  It is thus no surprise that the first pass at integration looks clunky.

The crux of the integration issue here is that there is no elegant way to map vertices to indices in this graph implementation.  However, there are a variety of other implementations of graphs that provide such a transformation very naturally.  Even if this is not the case, the only roadblock is the creation of a vertex to index mapping, which we could easily showcase.  Thus, I don't think that a repository of data structures & algorithms like this one should try to lock people in to the current implementation.  Part of being a software engineer is the ability to glue together the appropriate components in a pleasing way.

In short, I think that this post optimizes for simplicity in the wrong place; I think it is the data structures that should be straightforward.  I strongly believe that, in the long view, the composition of simple data structures will lead to more beautiful implementations of these algorithms.

src/algorithms/graph/kruskal/kruskal.js
```diff
@@ -33,28 +33,20 @@ export default function kruskal(graph) {
...
-    // Check if edge forms the cycle. If it does then skip it.
-    if (!disjointSet.inSameSet(currentEdge.startVertex, currentEdge.endVertex)) {
-      // Unite two subsets into one.
-      disjointSet.union(currentEdge.startVertex, currentEdge.endVertex);
+    const begIdx = vertices.indexOf(sortedEdges[edgeIndex].startVertex);
+    const endIdx = vertices.indexOf(sortedEdges[edgeIndex].endVertex);
 
-      // Add this edge to spanning tree.
-      minimumSpanningTree.addEdge(currentEdge);
+    if (disjointSet.find(begIdx) !== disjointSet.find(endIdx)) {
```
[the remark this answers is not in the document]
Bruce-Feldman: I agree the syntactic sugar of inSameSet names a common operation and makes the code more readeable.

I keep getting hung up on your assertion that the previous implementation is an object oriented one.  I do agree that it uses objects, but do not see the reasoning beyond this.

I think the conceptual cost of using objects is very large.  This cost, at the very least, includes the incorporation a significant amount of underlying javascript machinery that is not immediately obvious to a first time user.  It thus makes this algorithm impossible to reason about from a memory perspective, and difficult to reason about from a run time perspective.  This algorithm also becomes much less universal - it strongly depends on the quirks of the javascript language and, even worse, specific javascript run-time implementations.

src/data-structures/disjoint-set/DisjointSet.js
Bruce-Feldman: After experimenting with creating a disjoint set that takes in any object, I now find your objections to be conflicted.  With this implementation, you object to forcing the user to create a mapping from arbitrary inputs to indices.  However, with the previous implementation, you are fully on board with forcing the user to create a mapping from arbitrary objects to strings.

I fail to see how these are different except that in the previous implementation the requirements around the mapping function are not documented.  For instance, a mapping function from arbitrary inputs to unique objects will fail!  This was not at all obvious to me.  I am now very convinced that, even if you don't think that the previous implementation requires a change, the key function should be much more meticulously documented.

PR #114, opened 2018-07-26
Added running statistics

Bruce-Feldman:


trekhleb:
@Bruce-Feldman thank you for your PR.

I would like to ask you to make it more **focused** though. We need to have exact name of the algorithm or calculation you're trying to implement with its description in README.

Instead of just "statistics" link in main README it should be something like "Welford Algorithm - it deals with this and that...". If it is "Welford" then the class name need to be named something like "Welford" instead of "RunningStatistics". On contrary if you're implementing "RunningStatistics" than file should be name also "RunningStatistics.js" (the same as class name inside). The README file should have a description of what is "Welford Algorithm" or what is "Running Statistics".

PR #117, opened 2018-07-27
Update README.ko-KR.md

yeonjuan:
Update README.ko-KR.md for for Trapping Rain Water problem.


trekhleb:
Thank you for keeping Korean translation updated!
There is a conflict though. Could you please update your branch with the latest changes from https://github.com/trekhleb/javascript-algorithms master branch and **then** make a change in README?

The issue is that you have a file with underscore README.ko_KR.md but it has been renamed to README.ko-KR.md.

PR #121, opened 2018-07-30
Update LinkedList.js

seIncorp:
It is not necessary to use 'while' when user want delete head.
Beside this, you do not need check if head is null twice.


seIncorp:
I saw it and for that reason I wrote this issue:
Wrong description #122

PR #130, opened 2018-07-31
Add fibonnaci sequence closed form function

TheKarlo95:
Added function to find n-th fibonnaci sequence using closed form function. It is O(1) for numbers 1 to 75. If n is bigger than 75 difference in expect result and result this function gives is proportionally bigger because of rounding floating point numbers.


trekhleb:
Thank you for PR!

PR #136, opened 2018-08-03
fibonacci logn iterative implementation

nitros12:
This implements the `fib(n)` function that completes in `log(n)` number of iterations.
Found in SICP https://mitpress.mit.edu/sites/default/files/sicp/full-text/book/book-Z-H-11.html#%_thm_1.19


src/algorithms/math/fibonacci/fibonacciLogn.js
```diff
@@ -0,0 +1,35 @@
...
+  let count = n;
+
+  while (count > 0) {
+    if (count & 1) {
+      const nA = (b * q) + (a * q) + (a * p);
+      const nB = (b * p) + (a * q);
+      a = nA;
+      b = nB;
+      count -= 1;
+    } else {
```
Bruce-Feldman: You can remove this else if you want.

PR #142, opened 2018-08-05
Errors in LinkedList

Romvik:
I think it's just more like a typo, it should be a if statement


trekhleb:
@Romvik thank you for your PR. The tests are failing that means that if we change `while` to `if` then head deletion is starting to fail in case if there are several equal values at the head of the linked list. So I guess we still need to have `while` there.

Romvik:
Hm ok. I was finking the goal is to delete only one node at a time, because of the `return deletedNode;` in this case it will always return only one last node that was deleted.

PR #146, opened 2018-08-08
Update README.zh-CN.md and Add date

chinanf-boy:
I think Need to Add date :-O


trekhleb:
Thank you for the fixes, @chinanf-boy.
I don't think we need links to commits though. The changes that have been made to main README may be deducted from git history.

PR #152, opened 2018-08-11
Add countSetBits

tapasweni-pathak:
This adds implementation of counting set bits in a number using bitwise operators.


trekhleb:
@tapasweni-pathak Thank you for Pull Request! It is merged now!

PR #155, opened 2018-08-11
Add ifPowerOf2

tapasweni-pathak:
This adds implementation to check if a number is power of 2 using bitwise operators.


trekhleb:
Thank you for PR @tapasweni-pathak

PR #156, opened 2018-08-11
Add multiplyBy7

tapasweni-pathak:
This adds implementation of multiplying a number by 7 using bitwise operators.


trekhleb:
@tapasweni-pathak Thank you for PR!

We already have more general solution for bitwise multiplication in this repo. If you don't mind let's not include this specific one. But anyways thank you for your work!

Issue #108, opened 2018-07-24
Trie - remove words


workofart:
Removing words in a Trie seems equally important as adding words. It would be nice to have that.

PR #168, opened 2018-08-15
Add Trie Remove Words

damanm24:
Added removeWord function to the Trie class. See issue: [#108 ](https://github.com/trekhleb/javascript-algorithms/issues/108) for more info. To implement the removeWord function I also had to add a remove child function to the TrieNode class. Tests were added for both functions.


appleJax:
Hey @damanm24, I was going to just review and give feedback on your PR, but I figured it would be easier for me to put my thoughts into code and compare the results. I submitted an alternate PR #181. I would be happy to discuss the pros and cons of each of our approaches with you, @trekhleb, and whoever else is interested.

I like your recursive approach to deleting trieNodes, and it was thoughtful of you to make the recursive function private. I also like that you delegate the actual deletion of nodes to the TrieNode class.

One thing I changed in my implementation, is I naively call `trieNode.removeChild` from `Trie.deleteWord`, and the logic for deciding whether or not to actually delete the node is encapsulated within the `removeChild` method. I did this because as far as I can tell, a `trieNode` should never be deleted if it has children or if it has `isCompleteWord` set to `true`. The `deleteWord` function is therefore responsible for setting `isCompleteWord` to `false` for the last character in the word to be deleted.

Another difference in my implementation is I do not perform a `doesWordExist` check before executing the private recursive delete function. The reason for this is that `Trie.deleteWord` terminates early if a character is not found in the `Trie`.

Thanks for taking the time to submit this PR. I took a lot of inspiration from it. Let me know if you have any comments or questions. Hopefully between the two of us and whoever else chimes in, we can finalize an implementation that is fit to merge into master.

trekhleb:
@damanm24 thank you very much for your PR. I had two PRs on the same topic of removal from Trie: the current one and https://github.com/trekhleb/javascript-algorithms/pull/181. So I needed to choose one. I'll close the current PR in favor of https://github.com/trekhleb/javascript-algorithms/pull/181. But anyways thank you for your effort!

PR #171, opened 2018-08-15
choose pivot from median of three elements for quicksort

larskotthoff:
The method for choosing the pivot as implemented at the moment results in degenerate behaviour when the array is already sorted or reverse sorted. A better way is to choose the median of three elements (first, mid, last) as the pivot. See Robert Sedgewick. Implementing Quicksort Programs. Commun. ACM, 1(10):847–857, October 1978.

### DIFF
--- a/src/algorithms/sorting/quick-sort/QuickSort.js
+++ b/src/algorithms/sorting/quick-sort/QuickSort.js
@@ -14,13 +14,21 @@ export default class QuickSort extends Sort {
       return array;
     }
 
-    // Init left and right arrays.
+    // Init left, center, and right arrays.
     const leftArray = [];
     const rightArray = [];
+    const centerArray = [];
 
-    // Take the first element of array as a pivot.
-    const pivotElement = array.shift();
-    const centerArray = [pivotElement];
+    // Take the median element of first, mid, and last elements.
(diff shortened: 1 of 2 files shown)


larskotthoff:
Test failures seem to be unrelated? Not sure what's going on there.

orist22:
I think it is critical to choose appropriate pivot. If you fix the conflict, it would be really helpful!

orist22:
image
The code bot says that you exceeded the maximum size of array. I hope it would be helpful!

larskotthoff:
Should be fixed -- I wasn't choosing pivots the right way, and it turned out Kruskal's algorithm was broken...

RayJune:
> The method for choosing the pivot as implemented at the moment results in degenerate behaviour when the array is already sorted or reverse sorted. A better way is to choose the median of three elements (first, mid, last) as the pivot. See Robert Sedgewick. Implementing Quicksort Programs. Commun. ACM, 1(10):847–857, October 1978.

What a nice improvement.  @larskotthoff Thanks a lot.

The fact that this PR is not merged in is a loss for the repository.

lazarljubenovic:
The repo is mostly dead.

PR #173, opened 2018-08-20
Update README.md with license image

raineydavid:


trekhleb:
@raineydavid thank you for PR! But I think there is no need to do the accent on the license on top of the README (comparing to the tests and coverage metrics). MIT license badge is also presented in top GitHub native tab.

Issue #129, opened 2018-07-31
Longest common substring does not handle unicode properly


dubzzz:
It seems that the algorithm longestCommonSubstring does not handle unicode characters properly:

```js
longestCommonSubstr('𐌵𐌵**ABC', '𐌵𐌵--ABC') === '𐌵𐌵'
// whereas the longest one should be ABC (in terms of number of code points)

// Number of code points:
[...'𐌵𐌵'].length === 2
[...'ABC'].length === 3

// Number of "characters":
'𐌵𐌵'.length === 4
'ABC'.length === 3
```

You should maybe add a note on the algorithm regarding this. Basically the problem can occur whenever the strings contain characters outside the BMP range (ie code points greater than 0xffff).

Feel free to close the issue whenever you want. The aim was just to signal the problem is case you want to patch it in a way.

trekhleb:
@dubzzz thank you very much for reporting this!

JXWJS:
Never mind

trekhleb:
@dubzzz thank you for reporting the issue. It should be fixed now.

PR #176, opened 2018-08-22
fix longestCommonSubstring() to handle unicode characters (#129)

daddou-ma:
A have fixed the `longestCommonSubstring` to handle unicode characters by spliting the string into an array of one character strings.


daddou-ma:
**before:**
```
'𐌵𐌵**ABC'[0] // will result "�"
```
**After:**
```
Array.from('𐌵𐌵**ABC')[0] // will result "𐌵"
```

trekhleb:
@daddou-ma thank you for the fix!

daddou-ma:
@trekhleb You're welcome ^^

PR #177, opened 2018-08-25
Fix TrieNode.addChild so substrings get marked as complete words

appleJax:
## Fix for Trie.addWord (actual code changed is in TrieNode.addChild)

Currently when adding a substring of an existing word, the new word does not get marked as a complete word. Here's an example:

```js
const trie = new Trie();
trie.addWord('javascript');
trie.addWord('java');

trie.doesWordExist('java'); // returns false
```

This PR fixes the issue.


trekhleb:
Thank you for the fix @appleJax

PR #178, opened 2018-08-25
Update multiplyUnsigned.js

UjjwalAryal:


trekhleb:
Hi @UjjwalAryal ! Thank you for your PR. Normally each file may contain more than one author. This is relevant also for the file `multiplyUnsigned.js`. That is why contributors are not specified in the file comments themselves but rather we allow GitHub to track the authorship. You're may check the authorship here on this page https://github.com/trekhleb/javascript-algorithms/graphs/contributors. You are already mentioned there. Thank you for your contributions!

PR #181, opened 2018-08-27
Feature: Add Trie.deleteWord and TrieNode.removeChild

appleJax:
This PR adds the ability to delete words from the Trie data structure.

**NOTE - when deleting a word:**
- sets `trieNode.isCompleteWord` to `false` for the last character in the word
- deletes trieNodes in depth-first order (last character -> first character) 
- does NOT delete trieNodes if they have children
- does NOT delete trieNodes if `trieNode.isCompleteWord === true` (this is to prevent substring words from being deleted)

Example usage:
```js
const trie = new Trie();
trie.addWord('carpet');
trie.addWord('car');

trie.deleteWord('carpet');

trie.doesWordExist('carpet'); // returns false
trie.doesWordExist('car'); // returns true
```

Issue: #108
Refactored From PR: #168

### DIFF
--- a/src/data-structures/trie/Trie.js
+++ b/src/data-structures/trie/Trie.js
@@ -26,6 +26,35 @@ export default class Trie {
 
   /**
    * @param {string} word
+   * @return {Trie}
+   */
+  deleteWord(word) {
+    function depthFirstDelete(currentNode, charIndex) {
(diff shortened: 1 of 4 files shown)


trekhleb:
@appleJax thank you for pull request!

PR #184, opened 2018-08-28
clean up Heap implementation

appleJax:
In the Heap implementation, I noticed a few areas that could be cleaned up, including:
- typos
- MinHeap-specific comments
- minor refactoring for simplification


trekhleb:
Thank you for cleaning things up, @appleJax !

PR #185, opened 2018-08-28
Partial translation of Simplified Chinese

suifengtec:
Partial translation of Simplified Chinese


trekhleb:
Wow! Great work! Thank you @suifengtec

PR #189, opened 2018-08-29
Simplify dpMaximumSubarray

appleJax:
Alternate PR to #160 

**Changes:**
- Remove special case handling of array with all negative numbers
- Keep track of start/end indices and return a copy of `maxSubarray` at the end, rather than keeping a greedy copy of `maxSubarray`


trekhleb:
@appleJax thank you for simplification.

PR #194, opened 2018-08-31
Add LinkedList traversal and reverse implementations

hanhdt:
**Changes:**
- Add LinkedList traverse implementation
- Add LinkedList reverse implementations (reverseTraversal and reverse)
- Update their pseudocode

### DIFF
--- a/src/data-structures/linked-list/LinkedList.js
+++ b/src/data-structures/linked-list/LinkedList.js
@@ -214,16 +214,17 @@ export default class LinkedList {
    * @return {LinkedListNode[]}
    */
   traverse(callback = undefined) {
-    const traversedNodes = [];
+    if (typeof callback !== 'function') {
+      throw new TypeError(
+        `traverse method requires a callback function as an argument.\nArgument given: ${typeof callback}`
+      )
+    }
+
     let currentNode = this.head;
+    const traversedNodes = [];
 
     while (currentNode) {
-      // If callback is specified then try to call the callback.
-      if (callback) {
-        callback(currentNode);
-      } else {
-        traversedNodes.push(currentNode);
-      }
+      traversedNodes.push(callback(currentNode.value));
       currentNode = currentNode.next;
     }
 
--- a/src/data-structures/linked-list/LinkedList.js
+++ b/src/data-structures/linked-list/LinkedList.js
@@ -211,6 +211,7 @@ export default class LinkedList {
   /**
    * Traverse through all nodes of the list from head to tail
    * @param {*} callback
+   * @return {LinkedListNode[]}
    */
   traverse(callback = undefined) {
     const traversedNodes = [];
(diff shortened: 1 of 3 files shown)


hanhdt:
@trekhleb Could you share thoughts on this please?

src/data-structures/linked-list/LinkedList.js
```diff
@@ -207,4 +207,88 @@ export default class LinkedList {
   toString(callback) {
     return this.toArray().map(node => node.toString(callback)).toString();
   }
+
+  /**
+   * Traverse through all nodes of the list from head to tail
+   * @param {*} callback
+   * @return {LinkedListNode[]}
+   */
+  traverse(callback = undefined) {
```
appleJax: It might be better to throw an error if the callback argument is not of type `function`.  `LinkedList` already has a `toArray` method, so it's never valid to pass a non-function arg.

If you verify that the callback is valid, you can simplify the method. Something like:
```js
traverse(callback) {
  if (typeof callback !== 'function') {
    throw new TypeError(
      `traverse method requires a callback function as an argument.\nArgument given: ${typeof callback}`
    )
  }

  let currentNode = this.head;
  const traversedNodes = [];

  while (currentNode) {
    traversedNodes.push(callback(currentNode.value));
    currentNode = currentNode.next;
  }

  return traversedNodes;
}
```

src/data-structures/linked-list/LinkedList.js
```diff
@@ -207,4 +207,88 @@ export default class LinkedList {
...
+      currentNode = currentNode.next;
+    }
+
+    return traversedNodes;
+  }
+
+  /**
+   * The items in the list have been traversed in reverse order
+   */
+  reverseTraversal(callback = undefined) {
```
appleJax: This implementation of `reverseTraversal` has a running time of `O(n^2)`. It would be more straightforward and more performant (`O(n)`) to just define `reverseTraversal` as `traverse().reverse()`.

If you want to write a `reverseTraverse` function as an algorithm exercise, it might be better to define it recursively, like this:
```js
function reverseTraverse(node, callback) {
  if (typeof callback !== 'function') {
    throw new TypeError(
      `reverseTraverse method requires a callback function as an argument.\nArgument given: ${typeof callback}`
    )
  }

  if (!node) return [];

  return  reverseTraverse(
    node.next, callback
  ).concat(callback(node.value));
}
```

src/data-structures/linked-list/__test__/LinkedList.test.js
```diff
@@ -217,4 +217,71 @@ describe('LinkedList', () => {
...
+
+  it('should traverse through all nodes of the list from head to tail', () => {
+    const linkedList = new LinkedList();
+
+    linkedList
+      .append(1)
+      .append(2)
+      .append(3);
+
+    const traversedNodes = linkedList.traverse();
```
appleJax: `linkedList.traverse()` should be passed the callback. Also, it should be passed a callback that changes the values, to distinguish it from `toArray()`. e.g. `const callback = (x) => x * 2;`

src/data-structures/linked-list/__test__/LinkedList.test.js
```diff
@@ -217,4 +217,71 @@ describe('LinkedList', () => {
...
+    const traversedNodeValues = [];
+    traversedNodes.map(node => traversedNodeValues.push(node.value));
+
+    expect(traversedNodeValues).toEqual([1, 2, 3]);
+  });
+
+  it('should traverse through all nodes with callback', () => {
```
appleJax: If you adopt the other changes suggested, this test can be deleted.

hanhdt:
Hi @appleJax,
Could you help me review them again?
Thanks.

trekhleb:
Hi @hanhdt, thank you for PR! 
@appleJax thank you reviews as well!

I would agree that linked list traversals are worthy to be put in new `algorithms/linked-list` section. This way we will do the same is we do with trees and graphs which have their BFS and DFS traversals as a separate algorithms.

Meanwhile `reverse()` method may be a good fit for LinkedList class itself.

So I'm going to merge these changes for now and I'll move traversal codes to new section.

Thank you for your effort guys!

PR #201, opened 2018-09-04
Perform multiplication of any two integers positive or negative throu…

adityahiran:
Using bit manipulations for multiplying two signed integers


trekhleb:
Thank you for PR @adityahiran

PR #203, opened 2018-09-06
Improve JSDocs in Stack.js

ludvon:
The functions' comments were copied from Queue.js, but some words were not replaced.
I also made some changes to the wording for clarification.


trekhleb:
@ludvon thank you for the fixes!

PR #204, opened 2018-09-10
feat(bitwise): Function to check if a number is positive

adityahiran:
Bitwise operations used to determine if a number is positive


trekhleb:
@adityahiran thank you for PR.

PR #207, opened 2018-09-17
Update euclideanAlgorithm.js

vamshi9:
Method 2 is easy to evaluate or understand without using recursion stack!


trekhleb:
Thank you @vamshi9

PR #208, opened 2018-09-17
change '克努斯-莫里斯-普拉特算法' to 'KMP算法'

mcdongWang:
In China, we usually call it 'KMP算法', 
So I chose to use a free translation instead of a literal translation.
And I may submit content about Chinese translations more frequently recently.

PS: the editor fixed some problems of having extra spaces at the end of several lines.


trekhleb:
@mcdongWang thank you for PR!

PR #209, opened 2018-09-19
Create Q_matrix.js

vamshi9:
This algorithm works in O(logn) time and much more powerful than the above algorithms


vamshi9:
Someone, please check the indentations! Code is absolutely correct in algorithm and syntax.

mik-laj:
Can you run `eslint --fix` command to fix problems? Then there will be only 15 errors to fix. It should be easy to fix.

PR #214, opened 2018-09-23
Fix Stack pop comlexity to be O(1)

yavorski:
* By definition Stack push/pop time complexity should be O(1).
* Fix is applied by removing head instead of tail in pop method.
* Push method now do preprend instead of append.
* Fix consistency between toString and toArray methods.


trekhleb:
@yavorski thank you for your refactoring! Good job!

PR #215, opened 2018-09-27
Add iterative quick sort

yavorski:
Hello,

Iterative quick sort is a popular interview question, so i think it should be nice to have it here.

### DIFF
--- a/src/algorithms/sorting/quick-sort/QuickSortIterative.js
+++ b/src/algorithms/sorting/quick-sort/QuickSortIterative.js
@@ -1,0 +1,96 @@
+import Sort from '../Sort';
+import Stack from '../../../data-structures/stack/Stack';
+
+export default class QuickSortIterative extends Sort {
+  /**
+   * Iterative Quick Sort
+   *
+   * @param {*[]} originalArray - Not sorted array.
+   * @param {number} inputLowIndex
+   * @param {number} inputHighIndex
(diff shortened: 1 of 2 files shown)


src/algorithms/sorting/quick-sort/QuickSortIterative.js
```diff
@@ -0,0 +1,96 @@
...
+      // final sorted position.
+      swap(partitionIndex, highIndex);
+
+      return partitionIndex;
+    };
+
+    /**
+     * Replace recursion with auxiliary stack
+     */
+    const stack = new Stack();
```
appleJax: Nice addition! Just to nitpick, it might be a little cleaner to store arrays on the stack. That way, each item in the stack corresponds to a single iteration.
e.g.
```js
    let lowIndex = inputLowIndex;
    let highIndex = inputHighIndex;
    let partitionIndex;
    
    stack.push([ lowIndex, highIndex ]);

    while (!stack.isEmpty()) {
      [ lowIndex, highIndex ] = stack.pop();
      partitionIndex = partitionArray(lowIndex, highIndex);
      if (partitionIndex - 1 > lowIndex) {
        stack.push([ lowIndex, partitionIndex - 1 ]);
      }
      if (partitionIndex + 1 < highIndex) {
        stack.push([ partitionIndex + 1, highIndex ]);
      }
    }
```

PR #218, opened 2018-10-01
Fixed the translation in spanish

agusnavce:
Fixed some words of the spanish translation that weren't well written


trekhleb:
Thank you for PR, @agusnavce !

PR #221, opened 2018-10-08
Mutiple fixes in the spanish readme

jpvg10:
Fixing multiple spelling issues in the Spanish readme, and improving some vocabulary to make it more in line with the names you might find in Spanish algorithms and data structures literature.


trekhleb:
Thank you for your work @jpvg10 !

PR #222, opened 2018-10-09
Adding missing algorithms in the spanish readme

jpvg10:
When comparing the Spanish and English readmes, I noticed there were some algorithms missing in the spanish one.


trekhleb:
@jpvg10 thank you for PR!

PR #224, opened 2018-10-13
Properly detect min and max element in array

yavorski:


Dignity1988:
https://www.google.com.ua/search?client=ms-android-lenovo&ei=R8PBW9LmJISfsAHgzI3IDA&ins=false&q=array+starts+from&oq=array+starts+from&gs_l=mobile-gws-wiz-serp.3..0i203j0i22i30l4.2540.8958..9152...2.0..1.124.3835.34j9......0....1.......5..0j46j35i39j46i39j0i67j0i131j0i10j0i10i203.4mq7JgaAmBs  ... array starts from 1???..

lazarljubenovic:
`0` is wrong there indeed but why would you put `1` instead?

The correct starting values for the `reduce` way of calculating extrema are `Infinity` for the minimum and `-Infinity` for the maximum.

yavorski:
@lazarljubenovic not putting `1` at all, this is just the answer in this particular case with this specific array in mind. See diff for more details.

trekhleb:
@yavorski , good catch. Thank you for PR.

PR #226, opened 2018-10-16
BubbleSort: use Destructuring assignment to swap values

SidKwok:
Use Destructuring assignment to swap values.


trekhleb:
Thanks for PR @SidKwok

PR #227, opened 2018-10-16
Add prepend and fix bug

kiinlam:
Add prepend operation and fix some mistake in pseudocode.


trekhleb:
Thanks for PR @kiinlam

PR #240, opened 2018-10-28
add another solution

vipuljain08:


orist22:
image
Here is error message for user code.
It seems like there are strict rules for javascript code, but I think you can easily fix it by following error messages above!

PR #243, opened 2018-10-31
Changed push to spreadoperator

frankanneveld:
Improved performance


trekhleb:
@frankanneveld thank you for PR. But it brakes the tests. I guess we still need to push to arrays separately to support backtracking.

PR #244, opened 2018-11-02
Jest test - Example in README

deeper-x:
Hi there,
approaching this (wonderful) project, maybe user can find useful to start with a "working example" on jest-based playground in the README, so that he can stay focused on algorithm/data structure, immediately writing its own code in playground.
What do you think?

Thank you

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,12 +276,7 @@ describe('playground', () => {
 
 ```
 
-
-
-
-
-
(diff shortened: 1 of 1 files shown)


trekhleb:
@deeper-x thank you for suggestion! But I think we should not overload the front-page with the code examples.

PR #247, opened 2018-11-13
Fix DoublyLinkedList reverse

Herobs:
Previous link not update when reverse.


trekhleb:
@Herobs thank you for PR. I guess the similar changes have been already made in PR #257.

PR #250, opened 2018-11-18
update readme.zh-CN from readme

henrycjchen:


trekhleb:
Thank you @henrycjchen

PR #256, opened 2018-12-07
add astar algorithm

orist22:


GrosSacASac:
`arr.indexOf(endNode) !== -1`

 can be replaced with with 

`arr.includes(endNode) `

Issue #253, opened 2018-12-06
a bug in method reverse of doubly-linked-list


nsxz945:
https://github.com/trekhleb/javascript-algorithms/blob/2382225e033dc612e9688158212bee78a5b06f0f/src/data-structures/doubly-linked-list/DoublyLinkedList.js#L238-L260
the function omit changing target of one node's previous.

trekhleb:
Fixed in PR #257

PR #257, opened 2018-12-07
fixed doubly-linked-list reverse method omit change node's previous

jadestrong:
#253


trekhleb:
Thank you for the fix @jadestrong

PR #261, opened 2018-12-08
[ADD] Add huffman coding algorithm

yunji-yunji:
Huffman coding algorithm, one of the greedy algorithms, seems to be missing.

### DIFF
--- a/src/algorithms/uncategorized/huffman-coding/__test__/huffmanCoding.test.js
+++ b/src/algorithms/uncategorized/huffman-coding/__test__/huffmanCoding.test.js
@@ -1,0 +1,133 @@
+var Heap = function(fn) {
+    this.fn = fn || function(e) {
+      return e;
+    };
+    this.items = [];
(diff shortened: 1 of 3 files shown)


src/algorithms/uncategorized/huffman-coding/__test__/huffmanCoding.test.js
```diff
@@ -0,0 +1,133 @@
...
+    }
+};
+  
+var Huffman = {
+    // encode function
+    encode: function(data) {
+        var prob = {};
+        var tree = new Heap(function(e) {
+        return e[0];
+        });
```
lazarljubenovic: Something is wrong formatting here :thinking:

yunjiyunji: okay! I think it is better to use the arrow function.
Thank you! 😄

PR #262, opened 2018-12-09
Update the linke of trie to the chinese version

RequireSun:


trekhleb:
Thank you @RequireSun

PR #266, opened 2018-12-10
[FIX] Japanese Translation

Minki-Kim95:
日本人 to 日本語


trekhleb:
Oh, thanks for the fix!

PR #268, opened 2018-12-12
Add Fibonacci solution w/ recursion and memoization

mateusmlo:
This solution makes the recursive approach possible by storing computed results and uses some good functional and advanced concepts. I'm still new to GitHub, this is in fact my first PR ever, so sorry if there are mistakes.


mateusmlo:
Jokes on me the trailing spaces were not removed! 
await help;

jenilChristo:
This recursive implementation can posibly throw a `RangeException` on the `javascript call stack size`  for higher values of  input on `fibonacci()`.

andrzejsydor (approved):

PR #270, opened 2018-12-15
Added optimized solution for fibonacciNth problem

jenilChristo:
**##Description**
This PR adds a simple and  optimized solution for finding the Nth fibonacci problem, using backtracking in dynamic programming and other concepts like memoization and closures.

**Time complexity:**
Best case :  O(1) 
Worst case : O(n)


lazarljubenovic:
Which part exactly showcases a closure?

jenilChristo:
@lazarljubenovic  It's showcased where the function `fibonacciNth()` access  array `memoized` , which is defined outside its lexical scope via closure.

jenilChristo:
> Which part exactly showcases a closure?

Ok.so i found this was missing earlier as the param `memoize` was on global scope.just updated the code to showcase closure.Thanks for your comment :)

refer : https://github.com/jenilChristo/javascript-algorithms/blob/4c20687d0b7e9316076f84543e6b7f920b376838/src/algorithms/math/fibonacci/fibonacciNthBackTracking.js#L17

jenilChristo:
cc @trekhleb  Need your comments on this, please.

PR #282, opened 2019-01-04
[Fix] Capitalization issues of doubly linked list.

keisei77:
I was a little bit confused when I saw those capital letters, are those need to fix?


trekhleb:
Thank you for the fix, @keisei77

PR #290, opened 2019-01-14
Use increment instead of +=

arnavb:


arnavb:
Oh I wasn’t aware of that. Thanks for clearing it up.

Issue #265, opened 2018-12-09
QuickSort : Instead of taking pivot as 1st element, we should take the pivot as last element, due to time complexity


talentedandrew:
Instead of this
`const pivotElement = array.shift();` // O(n)

we should use 
`const pivotElement = array.pop();` // O(1)

When taking the pivot as last element 2 test cases are failing.
1. QuickSort › should do stable sorting
2. QuickSort › should visit NOT SORTED array element specified number of times

snehsumant:
Can I work on this issue ?

gkhedekar5758:
i would like to take this. can someone assign it to me please?

PR #300, opened 2019-02-08
switch shift to pop

henrycjchen:
according to my test, pop is faster than shift, especially when the array is very long.

### DIFF
--- a/src/algorithms/sorting/quick-sort/QuickSort.js
+++ b/src/algorithms/sorting/quick-sort/QuickSort.js
@@ -19,12 +19,12 @@ export default class QuickSort extends Sort {
     const rightArray = [];
 
     // Take the first element of array as a pivot.
-    const pivotElement = array.shift();
+    const pivotElement = array.pop();
     const centerArray = [pivotElement];
 
     // Split all array elements between left, center and right arrays.
     while (array.length) {
-      const currentElement = array.shift();
+      const currentElement = array.pop();
 
       // Call visiting callback.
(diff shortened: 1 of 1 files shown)


src/algorithms/sorting/quick-sort/QuickSort.js
```diff
@@ -19,12 +19,12 @@ export default class QuickSort extends Sort {
     const rightArray = [];
 
     // Take the first element of array as a pivot.
-    const pivotElement = array.shift();
+    const pivotElement = array.pop();
```
chandrunaik: shift and pop are used for different purposes. shift removes element from the front of the array and pop removes from the end. hence pop cannot used to get the first element in the array.

henrycjchen: in this sutiation, it no matter the element is taken from the front of the array or the end

mik-laj:
Related issues: https://github.com/trekhleb/javascript-algorithms/issues/265

src/algorithms/sorting/quick-sort/QuickSort.js
talentedandrew: @chandrunaik and @mrbeannnnn , in this case pop is better than shift because pop is O(1) (constant operation) where as shift() is O(n) .

henrycjchen: quick-sort should work well not metter the element is get from the top or from the end.
while in JS, pop is faster than shift, as talentedandrew said above.
so I think pop is better here.

PR #304, opened 2019-02-14
Fixes bugs related to unicode support in polynomial-hash functions

dubzzz:
This Pull Request fixes bugs related to unicode support in polynomial-hash functions:
- Method `charToNumber` of `PolynomialHash` was wrongly computed the code point value of a character: `codePointAt` is already aware of surrogate pairs (which is not the case of `charCodeAt`)
- Method `roll` of `PolynomialHash` ignored unicode characters totally
- `SimplePolynomialHash` was not aware of unicode

Originally I wanted to check how I could rewrite the existing for-loop based tests into something based on property based. Then it clearly appeared that the test can be simplified given the switch to property based: `roll(hash(XYZ), XYZ, YZA) === hash(YZA)` whatever the values of `X`, `Y`, `Z`, `A`.

The final version of the test can be read as follow:
For any `A` single character, `B` single character and `W` string,
`roll(hash(A + W), A + W, W + B) === hash(W + B)`

As a side note, I strongly believe that adding such kind of tests would highly help to detect such bugs easily. Up to now it has detected the following: #100, #101, #102, #110 and #129.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2263,6 +2263,16 @@
       "integrity": "sha1-lpGEQOMEGnpBT4xS48V06zw+HgU=",
       "dev": true
     },
+    "fast-check": {
+      "version": "1.10.1",
+      "resolved": "https://registry.npmjs.org/fast-check/-/fast-check-1.10.1.tgz",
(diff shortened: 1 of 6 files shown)


dubzzz:
@trekhleb Any updates? I can rebase the branch if needed

Issue #314, opened 2019-02-19
Could we have just one way of doing the swapping to have better consistency across the code 


cjkumaresh:
Like how we have here https://github.com/trekhleb/javascript-algorithms/blob/master/src/algorithms/sorting/bubble-sort/BubbleSort.js#L22 instead of
https://github.com/trekhleb/javascript-algorithms/blob/master/src/algorithms/sorting/shell-sort/ShellSort.js#L24

ktrongnhan:
The former makes use of ES6 destructuring whilst the latter does not. I think the question is whether ES6 is recommended in this project

mik-laj:
I feel that this project is trying to use modern language standards. I support the use of ES6 destructuring.

ekner:
Hi, we are four students working on this issue as part of a course in software development.

PR #356, opened 2019-05-26
Optimized code by removing tmp const

dhakerShivnarayan:
Removed tmp const and swapped elements using destructuring


mik-laj:
https://github.com/trekhleb/javascript-algorithms/issues/314
I think it is worth solving this problem globally.

PR #363, opened 2019-06-22
fix: return empty set when either is empty

GrosSacASac:


mik-laj:
Travis is sad ;_;

GrosSacASac:
Ping

Issue #378, opened 2019-08-05
Redundant condition


BnayaZil:
https://github.com/trekhleb/javascript-algorithms/blob/dc1047df725117c411959692a1ab538f65342b2e/src/algorithms/math/fibonacci/fibonacci.js#L13

This condition is redundant, the while condition covers the case where `n` equals `1`.

p.s- I will exploit this opportunity to say thank you for this repository.

BnayaZil:
Actually, if u want to be bolt prof u better check `n` for being greater than or equal to `1`.

ricardo93borges:
I created a PR that removes that condition: https://github.com/trekhleb/javascript-algorithms/pull/400

PR #379, opened 2019-08-07
:bug: Fix: #378 Removing Redundant Condition

ram-nad:
The code will now throw error if someone tries to index values less than 1.
Also code still works for n == 1 case.


ram-nad:
Also added the tests

PR #388, opened 2019-09-07
Russian version of the translation for README

awibox:
This will be very useful for Russian-speaking programmers


trekhleb:
Thanks for your contribution @awibox. However, the Russian version of the translation already exists. Feel free to update it if needed in a separate diff.

PR #400, opened 2019-10-03
Refactoring fibonacci algorithm

ricardo93borges:
Removed redundant condition


src/algorithms/math/fibonacci/fibonacci.js
```diff
@@ -10,10 +10,6 @@ export default function fibonacci(n) {
   let currentValue = 1;
   let previousValue = 0;
 
-  if (n === 1) {
```
nildakuzu: if you remove this condition, you should pop 1 from array

Issue #234, opened 2018-10-25
Add Bogosort


Shastel:

nas5w:
Is this something I could work on? I suspect the test cases might have to be customized for this algorithm (sorting a 20-element array using bogosort is impractical and can take infinite time).

PR #435, opened 2019-12-12
Add BogoSort

edumoreira1506:
Resolves issue #234


lazarljubenovic:
What's the point of this?

edumoreira1506:
The point is to add a new sort method to project :smile: See issue #234 .

lazarljubenovic:
But the sorting method is pointless. It's just dead code.

PR #438, opened 2019-12-18
Graph customization

Pierstoval:
* Allow graph edges with custom keys. This is important if we have two edges connected to the same vertices but having a different weight. We could have used the weight in the name, but as conceptually two edges could also have the same vertices and weight, I think it's a better idea to propose the user a way to make sure this edge case (pun intended) doesn't happen.
* Vertex can be converted to string, so I made sure that the `toString()` method is called, in case we have objects convertible to strings. The native `toString()` implementation might throw exceptions as expected.
* Edge key is now sure to be correctly converted to string too.
* We currently can't add an edge twice to a graph, I also added support for not adding a vertex twice.

### DIFF
--- a/src/data-structures/graph/Graph.js
+++ b/src/data-structures/graph/Graph.js
@@ -13,7 +13,13 @@ export default class Graph {
    * @returns {Graph}
    */
   addVertex(newVertex) {
-    this.vertices[newVertex.getKey()] = newVertex;
+    const key = newVertex.getKey();
+
+    if (this.vertices[key]) {
+      throw new Error('Vertex has already been added before');
+    }
+
+    this.vertices[key] = newVertex;
 
     return this;
   }
--- a/src/data-structures/graph/GraphEdge.js
+++ b/src/data-structures/graph/GraphEdge.js
@@ -2,22 +2,27 @@ export default class GraphEdge {
   /**
    * @param {GraphVertex} startVertex
    * @param {GraphVertex} endVertex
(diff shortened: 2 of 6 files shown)


moodmosaic:
I think this pull request looks reasonable except this part:

>We currently can't add an edge twice to a graph, I also added support for not adding a vertex twice.

What would be a valid scenario where adding a vertex twice does not make sense? :thinking:

Pierstoval:
Every vertex has a unique key, and vertex as objects can be referenced in the memory.

If you add a vertex, changing it afterwards should propagate the change to potential existing edge associated with it. If you just change the value in `this.vertices`, it will not be able to propagate to edges referencing this vertex.

The old behavior allowed to entirely **replace** a vertex, and if it was referenced somewhere, the reference was lost and concluded to a data corruption.

It's either this, or create a full-fledged propagation system, which is way more complex. I decided to go straightforward and disallow adding vertices twice :)

moodmosaic:
Interesting. Thank you for the explanation, @Pierstoval. Hoping that this gets merged, as it appears to be useful when using graphs for calculating currency arbitrage opportunities, for example.

Pierstoval:
Friendly ping @trekhleb : this improves graphs, and tests pass, what do you think? :)

Pierstoval:
Thanks! 🚀

lisagorewitdecker (commented):
Ok -

PR #470, opened 2020-03-16
Reduce traverseInOrder time complexity and space complexity

stephenkingsley:
In this case `traverseInOrder` use Recursive Approach. We can reduce the time complexity and the space complexity by using `Stack`.

#### Recursive Approach
Complexity Analysis
 * Time complexity : O(n). The time complexity is O(n) because the recursive function is T(n)=2⋅T(n/2)+1.
* Space complexity : The worst case space required is O(n), and in the average case it's O(logn)  where n is number of nodes.

#### Iterating method using Stack
Complexity Analysis
* Time complexity : O(n).
* Space complexity : O(n).


stephenkingsley:
@trekhleb CC! And could you please give my some advises?

stephenkingsley:
@trekhleb CC again please.

PR #471, opened 2020-03-20
i++

nainardev:


trekhleb:
In most cases in code there is a `i += 1` version. No really strong reason to change it to `i++`.

Issue #406, opened 2019-10-08
Add more cryptography  


adefirmanf:
Can these lists added in this repository as the cryptography section?  

- [ ] Caesar cipher
- [ ] Vigenere cipher
- [ ] Hill cipher

Thanks!

aesher9o1:
Adde Caesar Cipher #408

GledsonAfonso:
Doing Vigenere cipher.

GledsonAfonso:
Vigenere cipher done #414

rollingcomma:
Hill Cipher  [#424](https://github.com/trekhleb/javascript-algorithms/pull/424)

jprask:
Added rail fence cipher #515

PR #516, opened 2020-07-28
Add rail fence cipher

jprask:
Related to #406


trekhleb:
Thanks for implementing it, @jprask

PR #520, opened 2020-08-01
Add french translation for some chapters

boardens:
Translated Queue ("_File_") and Priority Queue ("_File de priorité_") chapters into french.
Added french language to lang redirection in their respective default readmes.


trekhleb:
Thanks for translations, @boardens!

PR #523, opened 2020-08-01
Create different_power_of2.js

2701gouravgoel:


src/algorithms/math/is-power-of-two/different_power_of2.js
```diff
@@ -0,0 +1,19 @@
+
+  // different power of 2 requiered so that the sum of different power of 2 equals to no..
```
stephencweiss, line 2: typo here - `requiered` (sic) should be `required`

2701gouravgoel: Can you please help me. What is sis why i am getting the error.

monpakh:
2701gouravgoel:xyz

monpakh:
2701gouravgoel:xyz

src/algorithms/math/is-power-of-two/different_power_of2.js
stephencweiss: "Sic" is meant to indicate that there's a typo without fixing it. 

Not sure on the error 😅

https://www.merriam-webster.com/words-at-play/sic-meaning-usage-editorial-citation

PR #532, opened 2020-08-27
added prime-factors algorithms under maths algorithms

pineapple45:
An algorithm to compute total distinct count of prime factors for a given natural number using the approximate Hardy-Ramanujan algorithm and its comparison with the original accurate algorithm.


gbumanzordev (approved):

trekhleb:
Thanks for the contribution, @pineapple45!

PR #541, opened 2020-09-18
chore: remove redundant code

beginWithJS:
I think the code I removed is not valid because the code after it and it should be removed.


beginWithJS:
mistake

PR #548, opened 2020-09-27
Add Mini-max sum

valarpirai:
Added Mini-max sum


src/algorithms/math/mini-max-sum/README.md
```diff
@@ -0,0 +1,11 @@
+# Mini-Max Sum
+
+Given an array of five positive integers, find the minimum and maximum values that 
+can be calculated by summing exactly four of the array integers. Then return the
+respective minimum and maximum values as a array of numbers
```
lazarljubenovic, line 5: This sounds like an interview question and not a general computer science algorithm. It should at the very least be expanded to `n` and `n-1`, although I still don't think it fits with the other algorithms in this repo.

lazarljubenovic (changes requested):

PR #552, opened 2020-10-01
Create AVL Tree

Varun2012:


src/algorithms/tree/AVL Tree
```diff
@@ -0,0 +1,288 @@
...
+
+struct node *deleteTree(struct node *tree)
+{
+    if(tree!=NULL)
+    {
+        deleteTree(tree–>left);
+        deleteTree(tree–>right);
+        free(tree);
+    }
+}
```
leandrocodes, line 288: Isn't that supposed to be a Javascript source code?

leandrocodes (changes requested):

trekhleb:
The files must be in JS format.

PR #571, opened 2020-10-19
Create taximete

kaashif29:


lazarljubenovic (changes requested):
An obvious Hacktober spam. We can close this.

PR #575, opened 2020-11-17
Add Horner's Method

MatheusBonavite:
Added Horner's Method into algorithms/math. Horner's method is a clever solution named after William George Horner, to evaluate polynomial functions given a certain point. I think it would be an interesting addition since it can be used in many hardware/software applications.

### DIFF
--- a/src/algorithms/math/horner-method/README.md
+++ b/src/algorithms/math/horner-method/README.md
@@ -2,7 +2,7 @@
 
 In mathematics, Horner's method (or Horner's scheme) is an algorithm for polynomial evaluation.
 With this method, it is possible to evaluate a polynomial with only n additions and n multiplications.
-Also, its storage requirements are only n times the number of bits of x.
+Hence, its storage requirements are n times the number of bits of x.
 
 Horner's method can be based on the following identity:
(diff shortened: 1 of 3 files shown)


src/algorithms/math/horner-method/README.md
```diff
@@ -0,0 +1,16 @@
+# Horner's Method
+
+In mathematics, Horner's method (or Horner's scheme) is an algorithm for polynomial evaluation.
+With this method, it is possible to evaluate a polynomial with only n additions and n multiplications.
+Also, its storage requirements are only n times the number of bits of x.
```
lazarljubenovic, line 5: I think the description would highly benefit from a comparison between Horner's scheme to a regular by-definition approach; it would also give meaning to the repeated words “only” in both sentences.

MatheusBonavite: Thank you, Lazar. I've added your suggestions. I removed the repeated "only" words and tried to make the text more clear. Also, gave comparison scenarios for a specific polynomial showing that Horner's rule would use a smaller number of mathematical operations when compared to a simple by-definition approach.

lazarljubenovic (approved):

trekhleb:
@MatheusBonavite, thanks for your contribution!

PR #581, opened 2020-11-27
Refactored merge sort to use array pointers instead of .shift()

austintheriot:
The previous implementation of merge sort used the array.shift() method to merge elements from leftArray and rightArray into the sorted array. This is a bad practice, because JavaScript's implementation of the Array.prototype.shift() method is, at worst, O(n) runtime complexity, where n is the length of the array being sorted (in order to move n elements of an array over by 1, all n elements of the array must typically be touched).

To work around this, I refactored merge sort to use array pointers instead. Now, after an element has been added to the sorted array, rather than delete it from the old array and shift the entire old array over, the array index pointer is just incremented to the right by one to avoid the old element. This will always require O(1) runtime complexity, which is a significant performance boost.

I also added two simple concat() statements to the returned sortedArray value, because the length of the left and right arrays can no longer be used to determine if an element is still "dangling" (needing to be added to the sorted array).

All tests passed after my changes.

[Quick Stack Overflow info on JavaScript array method runtime complexities](https://stackoverflow.com/questions/22614237/javascript-runtime-complexity-of-array-functions)

### DIFF
--- a/src/algorithms/sorting/merge-sort/MergeSort.js
+++ b/src/algorithms/sorting/merge-sort/MergeSort.js
@@ -1,4 +1,4 @@
-import Sort from '../Sort';
+import Sort from "../Sort";
 
 export default class MergeSort extends Sort {
   sort(originalArray) {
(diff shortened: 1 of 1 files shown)


trekhleb:
@austintheriot , thanks for the contribution!

PR #588, opened 2020-12-11
Added BMI algorithm

mrwolferinc:
I have decided to add an algorithm to calculate body mass index. I was originally going to add an algorithm to find the BMI of children and teens, but I was not able to do it.


lazarljubenovic:
This is just a formula. I don't see how this is a CS "algorithm".

mrwolferinc:
> This is just a formula. I don't see how this is a CS "algorithm".

Nevermind.

PR #589, opened 2020-12-12
Add a function powerOfTwoGreaterThanNumber

dilane3:
I have added one file to your repository, this file contains a function which take a number and returns the power of two which is directly greater than the given number. I have also added a test file and a README file.


dilane3:
How can I complete my checks ?, I remark that I need to do the **continuous integration** but how can I do it ?
Thanks

lazarljubenovic:
To be honest, this looks more like basic logarithm operations for a math class and less like a CS algorithm for this repo. Maybe if the code found the first `1` in the binary representation and cleared bits on the right?

dilane3:
> To be honest, this looks more like basic logarithm operations for a math class and less like a CS algorithm for this repo. Maybe if the code found the first `1` in the binary representation and cleared bits on the right?

dilane3:
For you I should not use the logarithm operation ? because this looks more like the math class ?

lazarljubenovic:
I'm not commenting on the code itself, I'm commenting on its inclusion in the project which has the usual bunch of CS algorithms. Calculating the "next power of two" is just applying the well-known logarithm function. In my opinion, it doesn't fit here. But I'm not the maintainer, just giving my two cents.

dilane3:
Ok. I understand that you say. But I wanted to do it because I have another contribution for this repository and which use it, then I have seen better to separate it.

This is my first pull request. I am a beginner, I would like to learn about git GitHub and JavaScript.

PR #593, opened 2020-12-16
Optimize Insertion Sort

AryanAhadinia:


JackyYin:
I don't get the advantage of this change.
You can swap the elements when compare the elements, why go through the list again?

AryanAhadinia:
> I don't get the advantage of this change.
> You can swap the elements when compare the elements, why go through the list again?

Consider we want to insert 5-th element (value = 2) in correct place in this list:
**1**, **3**, **4**, **5**, **6**, _2_, 8, 9, 7 --> **1**, _2_, **3**, **4**, **5**, **6**, 8, 9, 7

In pervious algorithm, we had to swap 4 times, (6, 2) and (5, 2) and (4, 2) and (3, 2). Since each swap needs assignment 2 times as least, we'll have variable assignment 8 times in minimum. In general, if we need swaps k times, we'll have 2 * k assignments.

But in new algorithm in same condition, we shift 3, 4, 5 and 6 then insert 2 in correct index. then we'll have k + 2 assignments which is better than 2 * k for k >= 2

JackyYin:
> > I don't get the advantage of this change.
> > You can swap the elements when compare the elements, why go through the list again?
> 
> Consider we want to insert 5-th element (value = 2) in correct place in this list:
> **1**, **3**, **4**, **5**, **6**, _2_, 8, 9, 7 --> **1**, _2_, **3**, **4**, **5**, **6**, 8, 9, 7
> 
> In pervious algorithm, we had to swap 4 times, (6, 2) and (5, 2) and (4, 2) and (3, 2). Since each swap needs assignment 2 times as least, we'll have variable assignment 8 times in minimum. In general, if we need swaps k times, we'll have 2 * k assignments.
> 
> But in new algorithm in same condition, we shift 3, 4, 5 and 6 then insert 2 in correct index. then we'll have k + 2 assignments which is better than 2 * k

@AryanAhadinia  Thanks for your explaination. 
How about this version to avoid go through the k elements again?
`  sort(originalArray) {
    const array = [...originalArray];

    // Go through all array elements...
    for (let i = 0; i < array.length; i += 1) {
      let currentIndex = i;
      let tmp = array[i];

      // Call visiting callback.
      this.callbacks.visitingCallback(array[i]);

      // Go and check if previous elements and greater then current one.
      // If this is the case then swap that elements.
      while (
        currentIndex > 0
        && this.comparator.lessThan(array[currentIndex], array[currentIndex - 1])
      ) {
        // Call visiting callback.
        this.callbacks.visitingCallback(array[currentIndex - 1]);

        // Swap the elements.
        array[currentIndex] = array[currentIndex - 1];

        // Shift current index left.
        currentIndex -= 1;
      }

      array[currentIndex] = tmp;
    }

    return array;
  }
`

AryanAhadinia:
Yes, it is correct I think, I'll work on this and I'll create a new pull request ASAP.

AryanAhadinia:
> > > I don't get the advantage of this change.
> > > You can swap the elements when compare the elements, why go through the list again?
> > 
> > 
> > Consider we want to insert 5-th element (value = 2) in correct place in this list:
> > **1**, **3**, **4**, **5**, **6**, _2_, 8, 9, 7 --> **1**, _2_, **3**, **4**, **5**, **6**, 8, 9, 7
> > In pervious algorithm, we had to swap 4 times, (6, 2) and (5, 2) and (4, 2) and (3, 2). Since each swap needs assignment 2 times as least, we'll have variable assignment 8 times in minimum. In general, if we need swaps k times, we'll have 2 * k assignments.
> > But in new algorithm in same condition, we shift 3, 4, 5 and 6 then insert 2 in correct index. then we'll have k + 2 assignments which is better than 2 * k
> 
> @AryanAhadinia Thanks for your explaination.
> How about this version to avoid the 2 * k assignments also avoid going through the k elements again?
> 
> ```
>   sort(originalArray) {
>     const array = [...originalArray];
> 
>     // Go through all array elements...
>     for (let i = 0; i < array.length; i += 1) {
>       let currentIndex = i;
>       let tmp = array[i];
> 
>       // Call visiting callback.
>       this.callbacks.visitingCallback(array[i]);
> 
>       // Go and check if previous elements and greater then current one.
>       // If this is the case then swap that elements.
>       while (
>         currentIndex > 0
>         && this.comparator.lessThan(array[currentIndex], array[currentIndex - 1])
>       ) {
>         // Call visiting callback.
>         this.callbacks.visitingCallback(array[currentIndex - 1]);
> 
>         // Swap the elements.
>         array[currentIndex] = array[currentIndex - 1];
> 
>         // Shift current index left.
>         currentIndex -= 1;
>       }
> 
>       array[currentIndex] = tmp;
>     }
> 
>     return array;
>   }
> ```

Actually, there is a little mistake in this code, I'll correct this and make PR in few seconds

AryanAhadinia:
Now it is optimized!

PR #595, opened 2020-12-18
Added kmeans clustering

avi09:
Hi Oleksii,
I've added the kmeans clustering with tests and README. Please tell if anything else is needed.
Thanks,
Avi


trekhleb:
@avi09, thanks for adding another ML algorithm!

PR #606, opened 2020-12-21
Add Tim Sort

Mohammadjafari80:
I implemented the Tim Sort. I tried to keep it in the same fashion as other sorting algorithms as much as possible. I even used the source code of insertion sort in order to make the code better to understand. The reason I used the reference method for insertion sort and merging over allocating a new memory is nothing but optimization.


GITHUBGUY12357999:
Hi, I got the CI test From the checks that I did. 


Proof:-- 


Error: We are currently unable to download the log. Please try again later.

PR #610, opened 2020-12-22
Optimize Insertion Sort

Mohammadjafari80:
I optimized the Insertion Sort by replacing the Linear Search method with the Binary Search method. in this way we have much fewer comparisons therefore the performance of the algorithm enhances significantly for a large amount of data;


yakmuswak:
That's great idea to move forward

yakmuswak:
Merge to create new ways

PR #617, opened 2020-12-28
Optimize Insertion Sort, update tests, & correct comments

austintheriot:
# Insertion Sort Clean Up

## Algorithm Change
The existing insertion sort implementation began by iterating from
0 until the end of the array, but it is only necessary to
iterate from 1 until the end of the array, since at the
0th index, there is nothing to compare to the left of
the element.

I also corrected the grammar/wording of the comments.

## Test Change
In order to complete this change, I also had to update the tests
to reflect the fact that the algorithm visits each index 1 less
time.

## Diagnostics
Passed all tests and linting.


trekhleb:
Thanks @austintheriot

PR #684, opened 2021-04-06
Chore/Refactoring for loop in PolynomialHash

ivensgustavo:
Change for loop to foreach loop.


trekhleb:
Package versions have been update in https://github.com/trekhleb/javascript-algorithms/pull/841

PR #703, opened 2021-04-29
add german translation for main readme

denizbinay:
I have left the English names of the algorithms in round brackets. Nobody uses the German terms in everyday life. They are used at universities, but never at work. But they help to understand. This is a good compromise. Unfortunately, there are also some that have never been given a German name. Then I didn't invent one either. I have tried to use only official terminology. That is, from specialist literature, Wikipedia, etc. To keep it consistent, a purely English name appears twice.


trekhleb:
Thanks for the translation, @denizbinay!

PR #812, opened 2021-12-12
Add factorialRecursiveTCO (Tail Call Optimized)

trainer2001:


kungfooman:
This is interesting, I wonder how you test this, since its dependent on the JS engine?

E.g. Chrome doesn't seem to support this: https://chromestatus.com/feature/5516876633341952

Debugger with stack trace:

image

With the code: 
```js
function factorialRecursiveTCO(number) {
  function fact(number, accumulator = 1) {
    console.log("fact", number);
    debugger;
    if (number < 2)
      return accumulator;
    else
      return fact(number - 1, accumulator * number);
  }
  console.log("factorialRecursiveTCO", number);
  return fact(number);
}
factorialRecursiveTCO(10)
```

PR #817, opened 2021-12-23
Null Reference Exception

Seymur:
When `n.next = tail` is true, we assign `n` to `tail` and `null` to `tail.next`, so `n.next` also becomes `null`. Then we assign `n.next.next` (because now `n.next` is `null`), we try to get `next` of `null`. That is why we should add an `else` case to check if `n.next` is not equal to `tail`.


GITHUBGUY12357999:
I will approve once I chk this.

GITHUBGUY12357999 (approved):
**approved**

PR #860, opened 2022-02-16
chore(docs): Typo in README

DaviDevMod:
The name of the language is "Italiano" (ending in -o), it becomes "Italiana" only when it has to match the genre of "lingua" ("language") which is feminine.

So you could translate "Italian language" as either "lingua italiana" or "italiano".

Pretty much the same happens, for example, in Spanish: you could say either "lengua española" or just "español".


DaviDevMod:
I closed the PR because all the READMEs need the same correction.

So I opened this other PR #862

PR #892, opened 2022-05-31
Better `isPowerOfTwo` algos and files

### LINKED ISSUE #889: `isPowerOfTwo` algos don't specify their ranges
Neither the README nor the src files explicitly say that the bitwise approach is only correct for 32bit ints. The way it's worded implies that the bitwise approach is always better, but the "naive" approach is the most mathematically correct, it even handles `NaN` and `Infinity` correctly! Why isn't there a clarification?

Rudxain:
Closes #889.

The bitwise approach isn't mathematically correct, because it only works for 32bit ints, therefore it should be exclusive to the `math/bits` directory.

I also added info about the supported input args for both the trial division approach and the bitwise approach


Rudxain:
Now the non-bitwise fn supports `BigInt`s

lisagorewitdecker:
> Closes #889.
> 
> 
> 
> The bitwise approach isn't mathematically correct, because it only works for 32bit ints, therefore it should be exclusive to the `math/bits` directory.
> 
> 
> 
> I also added info about the supported input args for both the trial division approach and the bitwise approach

lisagorewitdecker (commented):
Good stuff

m-coder567:
if pr closed why he is dosent merged this mentor pleas send it mail!

Issue #873, opened 2022-03-21
[BUG] Strongly connected components function breaks when adding edges on opposite directions


jackap:
Computing strongly connected components fails when adding edges in opposite direction. Minimal example:

```js      
        const vertexA = new GraphVertex('A');
        const vertexB = new GraphVertex('B');
        const vertexC = new GraphVertex('C');
        const vertexD = new GraphVertex('D');

        const edgeAB = new GraphEdge(vertexA, vertexB);
        const edgeBA = new GraphEdge(vertexB, vertexA);
        const edgeBC = new GraphEdge(vertexB, vertexC);
        const edgeCA = new GraphEdge(vertexC, vertexA);
        const edgeCD = new GraphEdge(vertexC, vertexD);

        const graph = new Graph(true);

        graph
            .addEdge(edgeAB)
            .addEdge(edgeBA)
            .addEdge(edgeBC)
            .addEdge(edgeCA)
            .addEdge(edgeCD);

        const components = stronglyConnectedComponents(graph);
```
Gives the following output: 

```
Uncaught Error: Edge has already been added before
    at Graph.addEdge (Graph.js:74:1)
    at Graph.js:149:1
```

itsamirhn:
I think this error is for reverse function in graph. Because as written [here](https://github.com/trekhleb/javascript-algorithms/blob/master/src/data-structures/graph/Graph.js#L139), edges reverse one by one and for example if we have both AB and BA edges, on reversing AB, BA will be exist and cause error.

I'm going to create a PR and fix this issue.

itsamirhn:
@trekhleb, Would you please take a look at this?

PR #916, opened 2022-07-30
Fix BUG in graph reverse method & Add needed tests

itsamirhn:
This PR fix the issue in #873.

The issue was in the old reverse method, edges reversed one by one and this was making issue for graphs with cycle of length 2. For example, when graph with only two edges AB and BA was reversing, on reversing AB, BA was exist and there will be crash.


itsamirhn:
Would you please check this PR? @trekhleb

itsamirhn:
Would you please check this PR? @trekhleb

PR #919, opened 2022-08-08
Create a factors.js

zhiwei8:
Create a factors.js to find the factor of a number which is input by the user.


src/algorithms/math/factors/factors.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * 
+ * Finds the factor of a number.
+ *
+ */
+const prompt = require('prompt-sync')({ sigint: true });
+
+dafault export function printFactors(x) {
```
kungfooman, line 8: Do you even test your code?

.github/workflows/CI.yml
```diff
@@ -1,4 +1,4 @@
-name: CI
+name: javascript CI
```
kungfooman, line 1: Why do you mess with CI?

Issue #821, opened 2021-12-29
Malay language for README


amirfakhrullah:
Hi, can I add a Malay version for README?

PR #927, opened 2022-08-25
Malay readme

j0pgrm:
malay version of read me #821


trekhleb:
Looks like most of the links are not working since they were translated automatically, i.e.:

```
[Nombor Kompleks](src/algoritma/matematik/nombor-kompleks) <--- here
```

KimbugweReagan:
the links are not working

PR #999, opened 2023-02-20
Changed the BinarySearch component from a function to an arrow function.

Ben-lad:
Changed the BinarySearch component from a function to an arrow function.

The original component was defined using the function keyword, but I updated it to use an arrow function instead. This change provides a more concise syntax and also avoids any issues with the binding of the "this" keyword.

I have tested the updated component and verified that it works as expected.

Thanks for considering this change!

### DIFF
--- a/src/algorithms/search/binary-search/binarySearch.js
+++ b/src/algorithms/search/binary-search/binarySearch.js
@@ -9,7 +9,7 @@ import Comparator from '../../../utils/comparator/Comparator';
  * @return {number}
  */
 
-export default function binarySearch(sortedArray, seekElement, comparatorCallback) {
+export default const binarySearch = (sortedArray, seekElement, comparatorCallback) => {
   // Let's create comparator from the comparatorCallback function.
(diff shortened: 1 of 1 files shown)


lazarljubenovic:
It's a top-level function. There's no issue with "this", nor is it more concise. Also it's not a component. 

Consider common sense before spamming multiple repositories with the same artificial-not-so-intelligent type of changes.

iamleomoko:
> It's a top-level function. There's no issue with "this", nor is it more concise. Also it's not a component.
> 
> Consider common sense before spamming multiple repositories with the same artificial-not-so-intelligent type of changes.

don't that hard on the guy, he might still be learning and just wants to contribute.

lazarljubenovic:
I looked at his profile; he had the same comments in other repos. It's a bot.

Ben-lad:
@lazarljubenovic I'm really sorry for that, if that make you angry. I'm not a bot just still learning as @iamleomoko said, and am looking for something small to change in the code. I'll delete this now 
So am really sorry once again.

@iamleomoko Thank you for saying that.

PR #1000, opened 2023-02-23
Big O and small O notations

ujaladev15:


andrzejsydor (approved):

amantank (commented):
This chart is quite complex and hard to read. My suggestion would be replacing this image with some easy to read and an image that is less complex like [this one.](https://github.com/trekhleb/javascript-algorithms/blob/8c5e5f4f0d5bb94439e02fe6ad6bb92b37a23834/assets/big-o-graph.png)

andrzejsydor (changes requested):
why such name of the file?

andrzejsydor (approved):

PR #1009, opened 2023-03-17
Update factorialRecursive.js by checking the factorial of negative numbers at run time

njiddasalifu:
added a clear check for the factorial eveluation of negative numbers to avoid wrong answer at runtime.

### DIFF
--- a/BACKERS.md
+++ b/BACKERS.md
@@ -8,21 +8,7 @@
 
 ## `O(n²)` Backers
 
-<table>
-  <tr>
-    <td align="center">
-      <a href="https://github.com/newrelic">
-        <img
-          src="https://avatars.githubusercontent.com/u/31739?s=200&v=4"
-          width="50"
-          height="50"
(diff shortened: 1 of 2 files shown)


lazarljubenovic:
It's bad practice to return a string or a number and then inspect it. You should throw an error for invalid input instead.

lazarljubenovic:
There's no such thing as "documen". Also, writing to the document is probably even worse than returning a string. You should throw an error instead.

lazarljubenovic:
You're throwing a string instead of an error now.

njiddasalifu:
take a look again.

olgunkaya (changes requested):
Do you really need that last number > 1 control ? You already guaranteed there the number is bigger than 1

njiddasalifu:
@olgunkaya ooh i see I will remove that. Actually the fisrt lines took care of that already.

Issue #1008, opened 2023-03-17
Factorial of a negative number does not exist


njiddasalifu:
This issue is addressing the fact that in the factorialrecursive implementation, it says if number is not greater than 1 then return 1. its not true for all cases cox fact of a neative is undefined.

njiddasalifu:
I have solved this already check PR for merged

HemanthPaila:
Hi there,

I want to contribute to this project and take on this issue as my first task. Could you please assign it to me? I am excited to work towards resolving it and will keep the team updated on my progress.

Thanks!

xyn22:
> I have solved this already check PR for merged

Please link your PR to this issue.

njiddasalifu:
@xyn22  you can find the PR link here https://github.com/trekhleb/javascript-algorithms/pull/1021

Leonardbrett:
> Hi there,
> 
> I want to contribute to this project and take on this issue as my first task. Could you please assign it to me? I am excited to work towards resolving it and will keep the team updated on my progress.
> 
> Thanks!

hello everyone I'm super excited I can't stop recommending her she help me a lot on trading I just made my withdraw of 22,108$ from my start of 5,800$ from Charlotte trading platform reach her on WhatsApp:+𝟭𝟰𝟳𝟱𝟮𝟰𝟮𝟲𝟯𝟯𝟴

lazarljubenovic:
But does she know what's the factorial of 0?

cynthiamarshal:
This issue pertains to the factorial recursive implementation. Currently, the implementation states that if the number is not greater than 1, it should return 1. However, this is not entirely accurate because the factorial of a negative number is undefined. [DGme](https://dgmeportal.live/)

utkarsh-shrivastav77:
Hi is this issue still open

jesicaakon:
That is correct. The factorial of a negative number does not exist. The factorial function is defined as the product of all positive integers less than or equal to a given number. Since negative numbers are not positive, they cannot be part of the product. Therefore, the factorial of a negative number is undefined.

Here are some examples of factorials of positive integers:

1! = 1
2! = 2
3! = 6
4! = 24
5! = 120
Here are some examples of factorials of negative integers:

(-1)! = undefined
(-2)! = undefined
(-3)! = undefined
(-4)! = undefined
(-5)! = undefined
You can also check factorial types of issues at [Employee Connection](https://ouremployeeconnection.com/)

[...]

PR #1011, opened 2023-04-02
throw-error-for-negative-factorial

Devesh35:
Issue: Factorial of a negative number does not exist #1008
Added condition to throw an error for negative numbers in factorial. Also added test cases to validate the results.


tanvirstreame (approved):

Devesh35 (commented):
Are any changes required.?

Issue #1030, opened 2023-04-16
I don't see kadane's algorithm for maximum subarray sum 


sundaram123krishnan:

sundaram123krishnan:
Have already added

BamaCharanChhandogi:
So, why this issue is open?

lazarljubenovic:
Because the PR isn't merged yet.

mithleshgupta:
I Would like to provide kandane algorithm in js

PR #1031, opened 2023-04-16
Added implementatin of kadane's algorithm

sundaram123krishnan:
This change attempts to resolve issue #1030.
Only included the implementation of the algorithm, let me know whether i should continue to contribute or not.
Signed-off-by : Sundaram krishnan <krishnsundaram@gmail.com>

### DIFF
--- a/src/algorithms/uncategorized/kadane's algorithm/kadane_algo.js
+++ b/src/algorithms/uncategorized/kadane's algorithm/kadane_algo.js
@@ -1,0 +1,18 @@
+/**
+ * Simplest Kadane's algorithm implementation.
+ *
+ * @param {*[]} array
+ * @return {number[]}
+ */
+
+export default function kadane() {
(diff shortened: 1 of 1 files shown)


src/algorithms/uncategorized/kadane's algorithm/kadane_algo.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * Simplest Kadane's algorithm implementation.
+ *
+ * @param {*[]} array
+ * @return {number[]}
+ */
+
+export default function kadane() {
+  let current_sum = array[0];
```
andrzejsydor, line 9: what is `array`?

sundaram123krishnan: Thanks for pointing out , i forgot to add parameter to function.

PR #1035, opened 2023-05-02
Added Linear congruential generator

ghost:
Adding a simple implementation of a linear congruential generator for generating n pseudorandom numbers.


AlgorithmicNonsense:
Might as well use a js generator function to save the state instead of n numbers beginning with seed.

PR #1037, opened 2023-05-08
addressed issue 1033, changed while to if in LinkedList and add new b…

### LINKED ISSUE #1033: Small code errors for the LinkedList and HashTable data structures
Hello!

As I went through the code for these data structures, I noticed two small errors:

1) inside the `delete()` method of the `LinkedList` data structure, there is a line of code that says: `while (this.head && this.compare.equal(this.head.value, value))`. However, instead of a `while` it should be an `if` statement.

2) inside the `constructor()` of the `HashTable` data structure, we initialize an array in the following way: `this.buckets = Array(hashTableSize).fill(null).map(() => new LinkedList());`. However, the `Array()` constructor should have the `new` keyword in front of it.

If you guys feel that these are important fixes to you and if you allow me to help, I would be more than glad to contribute by creating a pull request with the changes made.

tmichell13:
This change resolves #1033 by changing the 'while' statement in src/data-structures/linked-list/LinkedLists.js delete() method to an 'if', and adding the 'new' keyword before initializing an Array in src/data-structures/hash-table/HashTable.js constructor() method.

### DIFF
--- a/src/data-structures/hash-table/HashTable.js
+++ b/src/data-structures/hash-table/HashTable.js
@@ -12,7 +12,7 @@ export default class HashTable {
    */
   constructor(hashTableSize = defaultHashTableSize) {
     // Create hash table of certain size and fill each bucket with empty linked list.
-    this.buckets = Array(hashTableSize).fill(null).map(() => new LinkedList());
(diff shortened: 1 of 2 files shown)


Dawood3108:
.

PR #1050, opened 2023-07-06
Add Binary Search algorithm for finding Longest Increasing Subsequence

AbdulShaikz:
Implemented a binary search-based approach to find the longest increasing subsequence in an array.


AmanKumarVerma11 (approved):
Overall, this looks like a solid implementation of the binary search approach for the longest-increasing subsequence problem in JavaScript. The logic is correct and the code is clean and easy to follow. Some minor improvements with adding more comments would make it easier to understand for someone new to the algorithm.

PR #1055, opened 2023-07-30
fix(fastPowering): check when number is raided to power of one

jaw0r3k:
The function didnt check when power is one so unnecessarily was going though multiplying base by ones

### DIFF
--- a/src/algorithms/math/fast-powering/fastPowering.js
+++ b/src/algorithms/math/fast-powering/fastPowering.js
@@ -9,6 +9,11 @@
  * @return {number}
  */
 export default function fastPowering(base, power) {
+  if (power === 1) {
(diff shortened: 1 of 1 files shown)


chetannada (approved):
The solution by @jaw0r3k is right it's fastest then current solution.

Check the Screen shot for the proof
fastPowering (version 2) - JavaScript benchmark at JSBench me

Check Live Solution - [Click Here](https://jsbench.me/srlll6li1v/2)

lazarljubenovic:
This type of benchmark is usually not a good measurement for practical usage. More importantly, this type of "optimization" is not the goal of this repository -- it's about understanding data structures and algorithms, not squeezing out every nanosecond of the engine.

PR #1062, opened 2023-08-26
Added a recursive code for finding the fibonacci series

Shashi3k:
Hey, I added a recursive code for the fibonacci problem. Hope you merge it

### DIFF
--- a/src/algorithms/math/fibonacci/fibonacci.js
+++ b/src/algorithms/math/fibonacci/fibonacci.js
@@ -27,3 +27,16 @@ export default function fibonacci(n) {
 
   return fibSequence;
 }
+/**
+* Return a fibonacci series using recursion
+* Time Complexity -- O(2^N)
+* Space Complexity -- O(N)
+*/
+export default function fib(int n){
+  if(n==1 || n==0){
+    return n;
+  }
+  else{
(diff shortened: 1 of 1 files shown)


nihongeakira:
def fibonacci(n):
    if n <= 1:
        return n
    else:
        return fibonacci(n-1) + fibonacci(n-2)
You can use this function to get the nth Fibonacci number by calling fibonacci(n). However, keep in mind that this recursive approach has exponential time complexity. For large values of n, it will be slow. There are more efficient ways to compute Fibonacci numbers, such as using iterative methods or memoization.

If you'd like more information or improvements to the code, please let me know!

src/algorithms/math/fibonacci/fibonacci.js
```diff
@@ -27,3 +27,16 @@ export default function fibonacci(n) {
 
   return fibSequence;
 }
+/**
+* Return a fibonacci series using recursion
+* Time Complexity -- O(2^N)
+* Space Complexity -- O(N)
+*/
+export default function fib(int n){
```
andrzejsydor, line 35: please format the code

andrzejsydor (changes requested):

PR #1100, opened 2024-01-27
Update factorial.js

ihemanthm:
updated the factorial program to work with Negative Numbers also.


janoamaral (changes requested):
You should throw an error if the value is less or equal than 0

PR #1104, opened 2024-02-13
replaceChild method fixed

lpabinash:
replaceChild method fixed


trekhleb:
The `this.nodeComparator.equal` is used here for consistency. We may continue using it.

PR #1119, opened 2024-03-24
adding readme file explaining cocktailshakersort adding cocktailshake…

mohmmadAyesh:
Hi I added CocktailShakerSort Algorithm its like Bubble Sort Algorithm but it can be bidirectional in some cases it can achieve better than BubbleSort with less time Complexity on average and worst case its O(n^2) on best case its O(n) it take constant amount from Memory so its memory complexity is O(1) I Wrote A java script file for it following your pattern in writing these algorithm like extending class from Sort,etc Also make a test file CocktailShake.test.js to test it Also I added a Readme file that explaining it with references for further information Please Accept my PR.


mohmmadAyesh:
@trekhleb is that PR awaiting review or just I need to fix this test to pass please help thanks

PR #1124, opened 2024-04-12
fix(heap): this is undefined

ojj1123:
## What's the problem?
- `MaxHeapAdhoc` and `MinHeapAdhoc` constructor can get an array for filling the heap initially.
- And then the constructors execute this:
```js
this.heap.forEach(this.add);
```

- But It did occur this error:
  스크린샷 2024-04-13 01 19 15

- **The problem is that if passing `this.add` method literally, `this` in `add` method is `undefined`.**
- 
- This issue is confusing for the developers. **And your intend is that `MaxHeapAdhoc` and `MinHeapAdhoc` is used by developers easily**

https://github.com/trekhleb/javascript-algorithms/pull/1117

> This PR contains several minimalistic (by their functionalities and implementation) data structures like MinHeap, MaxHeap, and DisjointSet that don't have external dependencies and that are easy to copy-paste and use during the coding interview if allowed by the interviewer.

## How to fix
- Just use arrow function

```js
this.heap.forEach((value) => this.add(value));
```

### DIFF
--- a/src/data-structures/heap/MaxHeapAdhoc.js
+++ b/src/data-structures/heap/MaxHeapAdhoc.js
@@ -7,7 +7,7 @@
 class MaxHeapAdhoc {
   constructor(heap = []) {
     this.heap = [];
-    heap.forEach(this.add);
+    heap.forEach((value) => this.add(value));
   }
 
   add(num) {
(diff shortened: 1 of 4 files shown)


ojj1123:
@trekhleb
I was wondering if you could review this PR!
I think it is important since your intend could be broken up.

nitinjangra (approved):

Issue #1154, opened 2024-07-20
Least Common Multiple for a array of elements.


abhiphile:
It will be better to also have a `leastCommonMultiple` function for an array where I could pass an array of elements and get the result.
```js
export default function leastCommonMultiple(a, b) {
  return ((a === 0) || (b === 0)) ? 0 : Math.abs(a * b) / euclideanAlgorithm(a, b);
}
```

mukund1606:
Can I work on this one?

abhiphile:
Yes Sure !

mukund1606:
Created a PR for LCM on array of numbers

AnshDwivedi03:
Can i work on it

Vertabrae1:
1

Subhrajyoti842:
Can I do it ?

Subhrajyoti842:
Should I do it?

PR #1160, opened 2024-08-15
resolved 1154 issue - array-lcm

Madapati-Brungeshwar:
- Added the 'leastCommonMultipleArray' function to fix the issues: #1154.
- The function and the test code present at 'lcm-array' folder.


src/algorithms/math/lcm-array/_test_/lcm-array.test.js
```diff
@@ -0,0 +1,26 @@
+import leastCommonMultipleArray from '../lcm-array.js';
+
+describe('leastCommonMultiple', () => {
+  it('should find least common multiple', () => {
+    expect(() => leastCommonMultipleArray([])).toThrow(Error('Array is empty'));
```
Sparsh-katiyar, line 5: there is a one extra colon

Issue #2056, opened 2025-06-26
Potential Inefficiency and Missed Priority Updates in Dijkstra's Algorithm Implementation


Imran-imtiaz48:
In the Dijkstra algorithm, when a shorter path to a neighbor is found, the neighbor's priority in the priority queue should be updated regardless of whether it is already present in the queue.  
   In this code, the priority is only changed if queue.hasValue(neighbor) returns true. However, if a neighbor is not yet in the queue, it is added; but if it's already present, the code changes its priority.  
   This logic is correct as long as queue.changePriority works as intended, but it could be fragile if there are issues in the PriorityQueue implementation (such as not truly updating priorities or not handling duplicates). If the PriorityQueue does not remove duplicates, a neighbor may exist multiple times in the queue with different priorities.

PR #2064, opened 2025-08-01
dijkstra algo fixed ,duplicate files removed

rahul4581:
Solved #2056 
📝 Summary
This PR improves the Dijkstra algorithm implementation to ensure correct handling of priority updates for neighbors in the priority queue

🧠 Problem
In the current implementation, the neighbor’s priority in the queue is only updated if queue.hasValue(neighbor) returns true. This is logically sound only if the PriorityQueue:

Correctly changes the priority of an existing node, and

Does not contain duplicate entries.

If the PriorityQueue does not update priorities properly or allows duplicates, a node may exist multiple times in the queue with different costs — potentially breaking Dijkstra's guarantees.

✅ Fix
Explicitly ensure priority updates are handled correctly even if the node is already present in the queue.

Avoid potential issues caused by multiple entries of the same node with different priorities.


rahul4581:
just trying to contribute in open source projects  @lazarljubenovic bro

lazarljubenovic:
Huh, looks like it _is_ hard.

PR #2066, opened 2025-08-04
Linked list methods

AmirReza-alt:
# Add some usefull methods to LinkedList data structure.

List usually have some methods and properties that let user to give some information from the list like length.
I add some of this methods to the class to make it better.
This is a list of what methods i add:

- indexOf()
to find a index by it's value.

- at()
to find a value by it's index.

- print()
to print the entire list. example : 30 ==> 40 ==> 50 ==> null

- getSize()
to get the length of the linked list.
its better to define the property for this one but i prefere to don't touch the entire code and i just make a method for this one.

I hope it helps. This is my first contribution. If it's not a problem for you please tell me my mistakes to make myself better.😊

### DIFF
--- a/src/data-structures/linked-list/LinkedList.js
+++ b/src/data-structures/linked-list/LinkedList.js
@@ -329,4 +329,23 @@ export default class LinkedList {
 
     return currNode.value;
   }
+  
+  /**
+   * Print the entire list.
+   */
+  print() {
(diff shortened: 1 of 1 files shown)


anandfresh (approved):

dongthihuynhnhuxat81-sudo:
Thanks for sharing this! 👍

PR #2072, opened 2025-08-28
Manhattan Distance & Closest Pair Algorithm

Tarun2605:
### Summary

I have implemented Manhattan distance calculation functionality and also finding the closes pair of points when we are given a list of points in d dimensions
---

#### 1. Manhattan Distance Function

*  It will take two parameters a and b that will return the distance between the manhattan distance between them
* Time complexity: O(d) where d is the dimenstion

#### 2. Minimum Manhattan Distance Algorithm
* This algorithm takes a list of points of d dimension and returns the shortest manhattan distance among the pairs
* Optimum Time Complexity O(n log n · 2^d)  where  n is  the number of points given in the list and d is the dimension of the points


Tarun2605:
You mean the PR description? Sure.

PR #2128, opened 2026-03-28
Fix: Add missing word "structure" in Data Structures section

Allwell987:
Fixed a missing word in the Data Structures section.
"each data has its own trade-offs" corrected to
"each data structure has its own trade-offs".


andrzejsydor (changes requested):
I'm not sure if the PR has additional value.

PR #2179, opened 2026-06-06
fix(math): handle negative powers in fast powering algorithm

fauzan171:
## Problem

The `fastPowering` function enters infinite recursion when called with a negative power value.

```js
fastPowering(2, -1); // RangeError: Maximum call stack size exceeded
fastPowering(2, -2); // RangeError: Maximum call stack size exceeded
```

This happens because negative powers never hit the base case (`power === 0`) due to the floor division logic — `Math.floor(-1 / 2)` is `-1`, not `0`.

## Fix

Added handling for negative powers by computing the reciprocal before recursing:

```js
x^(-n) = 1 / x^n
```

This converts the negative power to a positive one and computes the reciprocal.

## Test Coverage

Added test cases for negative powers:
- `fastPowering(2, -1)` → `0.5`
- `fastPowering(2, -2)` → `0.25`
- `fastPowering(2, -3)` → `0.125`
- `fastPowering(4, -1)` → `0.25`
- `fastPowering(5, -2)` → `0.04`
- `fastPowering(10, -1)` → `0.1`

All existing tests continue to pass.

### DIFF
--- a/src/algorithms/math/fast-powering/__test__/fastPowering.test.js
+++ b/src/algorithms/math/fast-powering/__test__/fastPowering.test.js
@@ -1,6 +1,15 @@
 import fastPowering from '../fastPowering';
 
 describe('fastPowering', () => {
+  it('should compute negative powers correctly', () => {
+    expect(fastPowering(2, -1)).toBe(0.5);
+    expect(fastPowering(2, -2)).toBe(0.25);
+    expect(fastPowering(2, -3)).toBe(0.125);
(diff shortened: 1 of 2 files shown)


daltino (approved):
This PR addresses a clear bug by adding support for negative powers in the fast powering algorithm. The implementation is concise, follows the existing coding style, and is well-tested with sufficient test cases covering edge cases like different bases and negative exponents. All tests pass, and no unnecessary changes were introduced.

PR #2185, opened 2026-06-16
Use last element as QuickSort pivot

### LINKED ISSUE #265: QuickSort : Instead of taking pivot as 1st element, we should take the pivot as last element, due to time complexity

natecl:
## Summary
- Use the last element as the QuickSort pivot via `array.pop()` instead of using `array.shift()` from the front of the array.
- Preserve stable sorting by appending the popped pivot after equal elements that appeared before it.
- Update the expected visit count for the not-sorted fixture from `62` to `85`.

Closes #265.

## Verification
- `npx jest src/algorithms/sorting/quick-sort/__test__/QuickSort.test.js --runInBand`
- `npx eslint src/algorithms/sorting/quick-sort/QuickSort.js src/algorithms/sorting/quick-sort/__test__/QuickSort.test.js`

## Note
Related: #171 proposes a broader median-of-three QuickSort pivot change and also touches Kruskal. This PR is intentionally narrower and only addresses the `shift()` pivot cost described in #265.


daltino (commented):
The change to use the last element as the pivot in QuickSort seems reasonable, but it impacts the visiting count for the `NOT_SORTED_ARRAY`. The updated test ensures correctness, but the change to the visiting count requires verification to confirm it reflects the intended behavior. Also, double-check if these changes affect other test cases and the algorithm's performance.

PR #2197, opened 2026-07-19
fix: validate negative input in factorialRecursive function

REAPER2705:
- Add input validation to throw an error for negative numbers
- Update JSDoc to document the thrown error
- Prevent incorrect behavior where negative inputs would return 1


daltino (approved):
This PR ensures the `factorialRecursive` function properly validates negative input by throwing an error, which improves its reliability and aligns with mathematical conventions. The change is concise, follows the existing code style, and doesn't introduce issues based on the diff. However, confirm that tests for this validation exist or are added as required by the contribution guidelines.

PR #2198, opened 2026-08-05
fix(#2057): return ordered cycle path array instead of parent mapping in detectUndirectedCycle

### LINKED ISSUE #2057: Incorrect Cycle Path Construction in detectUndirectedCycle
In the function `detectUndirectedCycle`, the cycle path is being constructed as an object with vertex keys as properties and their parent vertices as values. This makes the returned `cycle` not a standard path array, but rather a mapping, which may not be intuitive or useful for consumers expecting an ordered path of vertices forming the cycle. A more conventional approach would be to return an **ordered array** of vertices representing the cycle, starting and ending at the same vertex.  
Additionally, if a cycle is detected, the function continues to traverse the graph until DFS completes, although traversal is blocked by `allowTraversal` once a cycle is set. However, early exit after detecting a cycle could improve efficiency.

**Suggested Fix:**  
- Change `cycle` to be an array that accumulates the ordered path of the cycle.
- Optionally, add a mechanism to exit the traversal immediately after the cycle is found.

gonzoblasco:
## Description

The `detectUndirectedCycle` function was returning an object mapping vertex keys to their parent vertices (e.g. `{ B: vertexC, C: vertexD, D: vertexE, E: vertexB }`), which is not a standard cycle representation and may not be intuitive for consumers expecting an ordered path.

This PR changes the return value to an ordered array of vertices forming the cycle, starting and ending at the same vertex (e.g. `[vertexB, vertexC, vertexD, vertexE, vertexB]`).

## Changes

- **detectUndirectedCycle.js**: Replaced the parent-mapping object with an ordered array built by walking the parent chain from the repeated vertex back to itself.
- **detectUndirectedCycle.test.js**: Updated the test expectation to match the new array format.

## Testing

```
Test Suites: 178 passed, 178 total
Tests:       587 passed, 587 total
```

Fixes #2057

### DIFF
--- a/src/algorithms/graph/detect-cycle/__test__/detectUndirectedCycle.test.js
+++ b/src/algorithms/graph/detect-cycle/__test__/detectUndirectedCycle.test.js
@@ -31,11 +31,12 @@ describe('detectUndirectedCycle', () => {
 
     graph.addEdge(edgeDE);
 
-    expect(detectUndirectedCycle(graph)).toEqual({
-      B: vertexC,
-      C: vertexD,
-      D: vertexE,
-      E: vertexB,
-    });
+    expect(detectUndirectedCycle(graph)).toEqual([
+      vertexB,
+      vertexC,
+      vertexD,
+      vertexE,
+      vertexB,
+    ]);
   });
 });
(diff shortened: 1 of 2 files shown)


daltino (commented):
Thanks for working on this! The changes look neat and well-structured.

m-coder567 (approved):
ALL WELL AND GOOD DEFIN GOOD JOB!

m-coder567:
Your works it's good and we all make it and solve it to geather that's why
this result comes out so all credit goes to you brother have take care nice
meeting you ❤️ I am just 17years old developer as web

m-coder567 (approved):
looks great and well defind structure so pr#2198 was going good lets get pushed it what you thinks gays!

